Issue #19, opened 2023-03-10
Normalize Markdown in `.pot` files


mgeisler:
When `mdbook-xgettext` extracts translatable text, it would be great if it could normalize the strings. This would make it possible for us to reformat the entire course without fearing that the translations get destroyed while doing so.

The normalization would take Markdown like this

~~~markdown
# This is a heading

This is another heading
=======================

A _little_
paragraph.

```rust,editable
fn main() {
    println!("Hello world!");
}
```

* First
* Second
~~~

and turn it into these messages in the `.pot` file:

* `"This is a heading"` ([atx heading](https://spec.commonmark.org/0.30/#atx-headings) is stripped)
* `"This is another heading"` ([setext heading](https://spec.commonmark.org/0.30/#setext-headings) is stripped)
* `"A _little_ paragraph."` (soft-wrapped lines are unfolded)
* `"fn main() {\n    println!("Hello world!");\n}"` (info string is stripped, we should instead use a [`#,` flag](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html))
* `"First"` (bullet point extracted individually)
* `"Second"`

Like in google/comprehensive-rust#318, we should do this in a step-by-step fashion and make sure to apply the transformations to the existing translations. It would also be good if we have a way to let translators update their not-yet-submitted translations.

djmitche:
I see the appeal, but I'm also wary of making it too easy to add too much code to a slide.  I think a wide-open `.rs` file in a text editor invites adding extra lines.

mgeisler:
> I don't think I'm going to get a chance to work on this issue soon.

That's alright, I'll look at it in the background and see what I can come up with.

mgeisler:
> @mgeisler I'm even hesitant, do you think I should wait a bit with [rust-unofficial/patterns#359](https://github.com/rust-unofficial/patterns/pull/359) and reinitialize the `de.po`-file when this feature is implemented?

I would not wait for this feature. My plan is to make it easy to move to a new version and keep the existing translations will continue to work. People (like yourself!) have put a lot of work into translations and we need to preserve this.

Concretely, I'm planning on providing a small program which people can run on the existing translations to normalize the Markdown found in them. So if the `messages.pot` file or a `xx.po` file contains

```po
#: src/welcome.md:10
msgid ""
"* Give you a comprehensive understanding of the Rust syntax and language.\n"
"* Enable you to modify existing programs and write new programs in Rust.\n"
[...]

PR #25, opened 2023-05-01
Implement fine-grained extraction of translatable text

mgeisler:
Before, we would extract text based on the byte offsets in the original document. As a consequence of this, the extracted text would look precisely like the original: the Markdown was copied directly from the original. In particular, text from a block quote would contain the leading ‘>’ characters and paragraphs in list items would contain leading whitespace.

Now, we instead extract text by grouping the Markdown parse events into those which should be translated and those who should be skipped. We use this in two ways:

- When extracting messages in ‘mdbook-xgettext’, we turn the translatable events back into Markdown. The structure of the document (headings, lists, block quotes, …) is no longer present in the extracted messages: only the text content itself it extracted.

- When translating, we replace the sequence of translatable events with the events from the translation. We do this while leaving the structure of the document unchanged.

The result of this is a much more robust system: editing one list item no longer impacts adjacent list items, moving a paragraph into a block quote no longer changes the paragraph.

As a side effect of how we turn events into messages, links are now all expanded. This makes the messages larger, but it removes a common source of errors where ‘[foo][1]’ would end up pointing to the wrong location if the reference link was updated.

Part of #19.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,256 +14,323 @@
 
 use mdbook::utils::new_cmark_parser;
 use pulldown_cmark::{Event, Tag};
-use std::ops::Range;
+use pulldown_cmark_to_cmark::{cmark_resume_with_options, Options, State};
 
-/// A translatable message.
-#[derive(PartialEq, Debug)]
-pub struct Message {
-    /// Line number where this message begins.
-    line: usize,
-
-    /// Span of the input text containing this message.
-    span: Range<usize>,
-}
-
-impl Message {
-    fn new(line: usize, span: Range<usize>) -> Self {
-        Self { line, span }
-    }
-
-    /// Get the text of this message, as a slice of the document from
-    /// which it was generated.
-    pub fn text<'doc>(&self, document: &'doc str) -> &'doc str {
-        &document[self.span.clone()]
-    }
-
-    /// Get the line number at which this message begins.
-    pub fn line_number(&self) -> usize {
-        self.line
-    }
-
-    /// Get the span of the source document from which this message is
-    /// drawn.
-    pub fn span(&self) -> Range<usize> {
-        self.span.clone()
-    }
-
-    /// Extend this message to the given offset.
-    fn extend(&mut self, to_end: usize) {
-        self.span.end = to_end;
-    }
-
-    /// Trim trailing newlines from this message.
-    fn trim_right(&mut self, document: &str) {
-        let trimmed_len = document[self.span.clone()].trim_end_matches('\n').len();
-        self.span.end = self.span.start + trimmed_len;
-    }
-}
-
-/// Accumulator for translatable messages based on input from the
-/// Markdown parser.
-struct MsgAccumulator<'a> {
-    /// The input document.
-    document: &'a str,
-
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -14,256 +14,321 @@
...
+    // Offsets of each newline in the input, used to calculate line
+    // numbers from byte offsets.
+    let offsets = text
+        .match_indices('\n')
+        .map(|(offset, _)| offset)
+        .collect::<Vec<_>>();
+
+    match state {
+        // If we're in a code block, we disable the normal parsing and
+        // return lines of text. This matches the behavior of hte
```
djmitche, line 55:
```suggestion
        // return lines of text. This matches the behavior of the
```

src/lib.rs
```diff
@@ -14,256 +14,321 @@
...
+        // parser in this case.
+        Some(state) if state.is_in_code_block => text
+            .split_inclusive('\n')
+            .enumerate()
+            .map(|(idx, line)| (idx + 1, Event::Text(line.into())))
+            .collect(),
+        // Otherwise, we parse the text line normal.
```
djmitche, line 62:
```suggestion
        // Otherwise, we parse the text line normally.
```

djmitche (approved):
Very nice! How will you handle migrating existing translations?

mgeisler:
> Very nice!

Thanks!

> How will you handle migrating existing translations?

My next step is to write a little normalization tool: it should be enough to go through a `.po` file and run both `msgid` and `msgstr` fields through the `extract_messages` function: hopefully we end up with `n` messages from the `msgid` and `n` messages from the `msgstr`. This way we have `n` new pairs. If we get a different number of messages, we can mark some (or all) of the pairs as fuzzy in the `.po` file.

src/lib.rs
```diff
@@ -14,256 +14,321 @@
...
+
+    // Block quotes and lists add padding to the state. This is
+    // reflected in the rendered Markdown. We want to capture the
+    // Markdown without the padding to remove the effect of these
+    // structural elements.
+    let state_without_padding = state.map(|state| State {
+        padding: Vec::new(),
+        ..state
+    });
+    cmark_resume_with_options(events, &mut markdown, state_without_padding, options).unwrap();
```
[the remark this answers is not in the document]
mgeisler: > Is the idea to return an accurate state (`new_state`) but return markdown rendered without the padding?

Yes, precisely! The padding is the `"> "` and list indents and I'm trying to avoid putting that into the `.po` file.

src/lib.rs
```diff
@@ -14,256 +14,321 @@
...
+    let events = group.iter().map(|(_, event)| event);
+    let mut markdown = String::new();
+    let options = Options {
+        code_block_token_count: 3,
+        list_token: '-',
+        emphasis_token: '_',
+        strong_token: "**",
+        ..Options::default()
+    };
+    let new_state = cmark_resume_with_options(
```
mgeisler, line 213:
```suggestion
    // Advance the true state, but throw away the rendered Markdown since it can contain padding.
    let new_state = cmark_resume_with_options(
```

PR #28, opened 2023-05-01
Add fuzzer for `group_events`

mgeisler:
The current version is pretty simple and the fuzz test is probably not necessary – however, a previous version was super hard to reason about and I kept losing events in subtle corner cases.

The fuzz test here helped me simplify the code and so I would like to keep it for future refactorings.


djmitche (approved):
Nice!

PR #34, opened 2023-06-16
Expand shortcut links like `[foo]`

mgeisler:
These links would normally end up as-is in the output text. This is a problem since it decouples them from the link definitions, which in turn makes it hard to translate since the message with the link is far removed from the message that defines the link.

Inlining everything ensures that each message is self-contained.


mgeisler:
I'll do this via another PR.

PR #61, opened 2023-08-23
Update USAGE.md

henrif75:
Remove '$' characters from shell commands to make it easier to copy and past.


mgeisler (approved):
Thanks, this aligns it with the the shell examples in Comprehensive Rust.

Issue #63, opened 2023-08-24
Add support for comments 


mgeisler:
We should teach `mdbook-xgettext` to react to comments in the Markdown. Comments could be used for:

- Marking the next paragraph or code block as non-translatable. This could save translators a lot of work if there are many code blocks which shouldn't be translated because they don't have any strings or comments.
-  Add translator comments. Gettext supports comments on each message and `xgettext` extracts them using a special syntax (which we should try to replicate).
- Add a message context. This is again s Gettext feature: the same message can appear multiple times, but with s different context each time.

dyoo:
I'm interested in working on this one.

dyoo:
Questions: 

I'm assuming we'll be looking at HTML comments?  

Do you already have some expectations on what the comment structure should look like for this operations? 

* To mark the next paragraph as non-translatable, for example, maybe: `<!-- SKIP -->`
* To add a translator comment, maybe `<!-- TRANSLATOR: ... ->`
* I don't know what a message context is quite yet: I will need to read more about what "s Getttext feature" means.  :)

I'm expecting that we'll want to expand the return type for `extract_messages()`, so that we can handle the translator comments or message contexts.  Does that match your expectations as well?

mgeisler:
> Questions:
> 
> I'm assuming we'll be looking at HTML comments?

Yes, precisely! They are our universal back-channel :slightly_smiling_face: 

> Do you already have some expectations on what the comment structure should look like for this operations?
> 
> * To mark the next paragraph as non-translatable, for example, maybe: `<!-- SKIP -->`

Yeah, but I would prefix it with `mdbook-xgettext:` to make it clear that this applies to this particular program:

````markdown
<!-- mdbook-xgettext: skip -->

```rust,editable
fn do_not_translate_me() {}
```
````

> * To add a translator comment, maybe `<!-- TRANSLATOR: ... ->`

This actually matches [this SO answer](https://stackoverflow.com/a/75947520/110204) which suggests calling `xgettext` as

```
xgettext --add-comments=TRANSLATORS
```

to extract C comments like this:

```c
// TRANSLATORS: A translator comment
printf(_("Hello world"));
```

As you might have guessed, our `mdbook-xgettext` plugin plays the role of `xgettext` from the normal Gettext suite. Since `xgettext` is configurable, we might want to make our keywords configurable as well. The configuration would be read from the `book.toml` file, similar to how we read `pot-file` today.

[...]

PR #72, opened 2023-09-05
Applies a workaround for parser bug dealing with inline comments.

dyoo:
This is some continuation of work for https://github.com/google/mdbook-i18n-helpers/issues/63 for implementing special behavior for HTML comments.  This PR splits off inline comments from content that immediately follows them, as a workaround  for https://github.com/raphlinus/pulldown-cmark/issues/712.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@
 //! how to use the supplied `mdbook` plugins.
 
 use polib::catalog::Catalog;
-use pulldown_cmark::{Event, LinkType, Tag};
+use pulldown_cmark::{CowStr, Event, LinkType, Tag};
 use pulldown_cmark_to_cmark::{cmark_resume_with_options, Options, State};
 use regex::Regex;
 use std::sync::OnceLock;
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @dyoo, thanks for the PR! We could patch this on our side, though I wonder if you've looked into fixing it upstream? I don't really know the internals of `pulldown_cmark` very well, but it would be super if you could spend an afternoon looking into it and perhaps improve this for everybody :slightly_smiling_face: 

I think the corner cases affected by this problem (in the skipping code) might be small enough that we don't have to work around them. I just tested with variations of the unit test:

```markdown
- foo
<!-- mdbook-xgettext: skip -->
- bar
- baz
```

works as expected and skips only `- bar`. I would be okay to say that

```markdown
- foo
- <!-- mdbook-xgettext: skip --> bar
- baz
```

just isn't supported right now. Especially because the consequences are small and easy to see: translators will see when they suddenly lose access to the rest of the list because both `bar` and `baz` are no longer extracted.

dyoo:
I did a little more analysis of the situation, and what is happening with the parser is following the CommonMark spec, so there is technically not a bug on their end.

The problematic situation is that, by starting with an HTML comment, we're creating an [HTML block](https://spec.commonmark.org/0.30/#html-block).  According to the specification, that block lasts till the end of the line.  The content of the HTML block is passed, unescaped, into the rendered content.

So the usage of block-level HTML comments is a little imperfect, and we have to be careful to be aware of this end-of-line behavior.  That does mean that if we're going to use `<!-- i18n ... ->` as our escape hatch, we do need to say something in usage notes to the effect that it must be the only thing on a line: otherwise, unintuitive things can happen.

I think we can close this bug.  I'll extract out some of the code refactorings I did to make the code a little nicer to read and do that as a separate PR.

mgeisler:
> So the usage of block-level HTML comments is a little imperfect, and we have to be careful to be aware of this end-of-line behavior. That does mean that if we're going to use `<!-- i18n ... ->` as our escape hatch, we do need to say something in usage notes to the effect that it must be the only thing on a line: otherwise, unintuitive things can happen.

Thanks for investigating this!

As you say, we should explain this eventually if/when people run into problems by trying to skip things in complex situations.

PR #75, opened 2023-09-07
Automatically ignore codeblocks that don't have lit-strings or line comments.

dyoo:
This tries to automatically omit codeblocks that don't have what looks like translatable content.  For the moment, we check whether the body of a codeblock contains a "\"" or "//" substring, which is a heuristic for checking for literal strings or line comments.

### DIFF
--- a/src/bin/mdbook-gettext.rs
+++ b/src/bin/mdbook-gettext.rs
@@ -205,10 +205,10 @@ mod tests {
     fn test_translate_code_block() {
         let catalog = create_catalog(&[(
             "```rust,editable\n\
-             fn foo() {\n\n    let x = 10;\n\n}\n\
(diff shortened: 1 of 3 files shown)


mgeisler (approved):
Very nice! I was somehow thinking that skipping a code block would use the new `mdbook-xgettext: skip` comments, but I guess there's no good reason to inject those comments just to take them out a split second later.

mgeisler:
I think we should consider making this configurable at a later point. I can imagine a few settings:

- skip all code blocks
- skip code blocks matching a given language (the info string following the ` ``` `)
- skip code blocks where the content matches a given set of regular expressions (such as `"\""` and `"//"` used above)

By default, code blocks would be extracted and we should probably only add the settings we need.

Let me merge this now, then we can make a new release and test it on Comprehensive Rust for a bit.

Issue #67, opened 2023-08-29
Support splitting the PO files into smaller units


mgeisler:
We currently work with monolithic file: `mdbook-xgettext` writes all messages to a single POT file (typically `messages.pot`) and `mdbook-gettext` read a single PO file as well. We've seen in [Comprehensive Rust](https://github.com/google/comprehensive-rust) how these files can grow very large: the PO files are around 800 kB and 24k lines in size.

It would be interesting to support splitting the POT file into smaller units. We could let the book outline decide the splitting and split on each top-level chapter. More elegantly, we could have a splitting depth: the current behavior would be splitting at depth 0, depth 1 would split at top-level chapters, depth 2 would go one level further, etc.

This could in principle be implemented today by scripting calls to `msggrep --location` (to split the POT file into smaller units) and `msgcat` (to compile multiple PO files). However, a native solution will likely be cleaner overall. The splitting with `msggrep` and combining with `msgcat` suggests an implementation strategy: keep everything like today, and split the `Catalog` into units before saving them in `mdbook-xgettext`. Similarly, create a large `Catalog` from many smaller ones in `mdbook-gettext`.

antoniolinhart:
Hey,

This bug looks interesting! Would I be able to pick it up? 

Is `Catalog` a data type for mdbook or just an unofficial term for a POT file? I tried to look up `Catalog data type` but I didn't see it associated with mdbook.

Thank you!

mgeisler:
Hi @antoniolinhart, I'm glad you're interested in it!

The [`Catalog` type](https://docs.rs/polib/latest/polib/catalog/struct.Catalog.html) is from polib, the library we use to read/write the PO files.

antoniolinhart:
> The recommended workflow would be to create a fork on GitHub and push to a branch in that repository. That will be your own playground which you have complete control over. You can create and delete branches in your fork as you like, without any danger of making permanent changes in this repository.

Wow! This makes a lot more sense now... I will do this for my next PRs. The process is easier than I thought. Thanks!!

mgeisler:
> Wow! This makes a lot more sense now... I will do this for my next PRs. The process is easier than I thought. Thanks!!

You can definitely get write access to this repository as well — I just wanted to point out that it's not strictly necessary most of the time :slightly_smiling_face: Your PRs will automatically start running tests after the first one is merged.

mgeisler:
[...]

PR #77, opened 2023-09-13
Split POT file depending on specified depth

antoniolinhart:
Hey all! This is a PR in relation to #67 to split the pot file depending on
the specified depth. Another PR will be submitted to change `gettext`
to merge a directory of `pot` files.

- Add an input param to `xgettext` to take in an integer for the depth to
  split the POT file. Remove the input-file param
- Mimics the same structure as the SUMMARY file for the Book
- Revise the helper fn to create a render ctx for tests. Now, it can
  create on-the-fly directories if the markdown file exists in a
sub-directory
- Test the new split catalog function with different possible inputs

### DIFF
--- a/i18n-helpers/src/bin/mdbook-xgettext.rs
+++ b/i18n-helpers/src/bin/mdbook-xgettext.rs
@@ -17,30 +17,30 @@
 //! This program works like `xgettext`, meaning it will extract
 //! translatable strings from your book. The strings are saved in a
 //! GNU Gettext `messages.pot` file in your build directory (typically
-//! `po/messages.pot`).
+//! `po/messages.pot`). When the `depth` parameter is included, a new
+//! directory will contain the template files split based on the tiers
+//! of Chapter nesting.
 
-use anyhow::{anyhow, Context};
+use anyhow::Context;
 use mdbook::renderer::RenderContext;
-use mdbook_i18n_helpers::xgettext::create_catalog;
+use mdbook_i18n_helpers::xgettext::create_catalogs;
 use std::{fs, io};
 
 fn main() -> anyhow::Result<()> {
     let ctx = RenderContext::from_json(&mut io::stdin()).context("Parsing stdin")?;
-    let cfg = ctx
-        .config
-        .get_renderer("xgettext")
-        .ok_or_else(|| anyhow!("Could not read output.xgettext configuration"))?;
-    let path = cfg
-        .get("pot-file")
-        .ok_or_else(|| anyhow!("Missing output.xgettext.pot-file config value"))?
-        .as_str()
-        .ok_or_else(|| anyhow!("Expected a string for output.xgettext.pot-file"))?;
     fs::create_dir_all(&ctx.destination)
         .with_context(|| format!("Could not create {}", ctx.destination.display()))?;
-    let output_path = ctx.destination.join(path);
-    let catalog = create_catalog(&ctx, std::fs::read_to_string).context("Extracting messages")?;
-    polib::po_file::write(&catalog, &output_path)
-        .with_context(|| format!("Writing messages to {}", output_path.display()))?;
+    let catalogs = create_catalogs(&ctx, std::fs::read_to_string).context("Extracting messages")?;
+
+    // Create a template file for each entry with the content from the respective catalog.
+    for (file_path, catalog) in catalogs {
+        let directory_path = file_path.parent().unwrap();
+        fs::create_dir_all(directory_path)
+            .with_context(|| format!("Could not create {}", directory_path.display()))?;
+
+        polib::po_file::write(&catalog, &file_path)
+            .with_context(|| format!("Writing messages to {}", file_path.display()))?;
+    }
 
     Ok(())
 }
--- a/i18n-helpers/src/xgettext.rs
+++ b/i18n-helpers/src/xgettext.rs
@@ -14,12 +14,13 @@
 
 //! This file contains main logic used by the binary `mdbook-xgettext`.
 
+use std::collections::HashMap;
 use std::{io, path};
 
 use super::{extract_events, extract_messages, reconstruct_markdown, wrap_sources};
 use anyhow::{anyhow, Context};
 use mdbook::renderer::RenderContext;
-use mdbook::BookItem;
+use mdbook::{book, BookItem};
 use polib::catalog::Catalog;
 use polib::message::{Message, MessageMutView, MessageView};
 use polib::metadata::CatalogMetadata;
@@ -87,16 +88,8 @@ fn dedup_sources(catalog: &mut Catalog) {
     }
 }
 
-/// Build catalog from RenderContext
-///
-/// # Arguments
-
-/// * `ctx` - RenderContext from mdbook library
-/// * `summary_reader` - A closure which reads summary at given path
-pub fn create_catalog<F>(ctx: &RenderContext, summary_reader: F) -> anyhow::Result<Catalog>
-where
-    F: Fn(path::PathBuf) -> io::Result<String>,
-{
+/// Build CatalogMetadata from RenderContext
+fn generate_catalog_metadata(ctx: &RenderContext) -> CatalogMetadata {
     let mut metadata = CatalogMetadata::new();
     if let Some(title) = &ctx.config.book.title {
         metadata.project_id_version = String::from(title);
@@ -109,6 +102,21 @@ where
     metadata.mime_version = String::from("1.0");
     metadata.content_type = String::from("text/plain; charset=UTF-8");
     metadata.content_transfer_encoding = String::from("8bit");
+    metadata
+}
+
+/// Build catalog from RenderContext
+///
+/// The `summary_reader` is a closure which should return the
+/// `SUMMARY.md` found at the given path.
+pub fn create_catalogs<F>(
+    ctx: &RenderContext,
+    summary_reader: F,
+) -> anyhow::Result<HashMap<path::PathBuf, Catalog>>
+where
+    F: Fn(path::PathBuf) -> io::Result<String>,
+{
+    let metadata = generate_catalog_metadata(ctx);
     let mut catalog = Catalog::new(metadata);
 
     // The line number granularity: we default to 1, but it can be
@@ -147,24 +155,181 @@ where
         );
     }
 
+    let mut catalogs = HashMap::new();
+
+    // The depth on which to split the output file. The default is to
+    // include all messages into a single POT file (depth == 0).
+    // Greater values will split POT files, digging into the
+    // sub-chapters within each chapter.
+    let depth = match ctx
+        .config
+        .get_renderer("xgettext")
+        .and_then(|cfg| cfg.get("depth"))
+    {
+        None => 0,
+        Some(value) => value
+            .as_integer()
+            .and_then(|i| (i >= 0).then_some(i as usize))
+            .ok_or_else(|| anyhow!("Expected an unsigned integer for output.xgettext.depth"))?,
+    };
+
+    // The catalog from the summary data will exist in the single pot
+    // file for a depth of 0, will exist in a top-level separate
+    // `summary.pot` file for a depth of 1, or exist within in a
+    // `summary.pot` file within the default directory for chapters
+    // without a corresponding part title.
+    let mut current_top_level = "summary".to_owned();
+    let mut summary_destination = match depth {
+        0 => path::PathBuf::from("messages"),
+        1 => path::PathBuf::from("summary"),
+        _ => path::PathBuf::from(&current_top_level).join("summary"),
+    };
+    let _: bool = summary_destination.set_extension("pot");
+    catalogs.insert(summary_destination, catalog);
+
     // Next, we add the chapter contents.
-    for item in ctx.book.iter() {
-        if let BookItem::Chapter(chapter) = item {
+    for item in &ctx.book.sections {
+        if let BookItem::PartTitle(title) = item {
+            // Iterating through the book in section-order, the
+            // PartTitle represents the 'section' that each chapter
+            // exists within.
+            current_top_level = slug(title);
+        } else if let BookItem::Chapter(chapter) = item {
             let path = match &chapter.path {
                 Some(path) => ctx.config.book.src.join(path),
                 None => continue,
             };
+            let directory = match depth {
+                0 => path::PathBuf::from("messages"),
+                1 => path::PathBuf::from(current_top_level.clone()),
+                // The current chapter is already at depth 2, so
+                // append the chapter's name for depths greater than
+                // 1.
+                _ => path::PathBuf::from(current_top_level.clone()).join(slug(&chapter.name)),
+            };
+
+            // Add the (destination, catalog) to the map if it doesn't
+            // yet exist, so messages can be appended to the catalog.
+            let mut destination = directory.clone();
+            let _: bool = destination.set_extension("pot");
+            let catalog = catalogs
+                .entry(destination.clone())
+                .or_insert_with(|| Catalog::new(generate_catalog_metadata(ctx)));
+
             for (lineno, extracted) in extract_messages(&chapter.content) {
                 let msgid = extracted.message;
                 let source = build_source(&path, lineno, granularity);
-                add_message(&mut catalog, &msgid, &source, &extracted.comment);
+                add_message(catalog, &strip_link(&msgid), &source, &extracted.comment);
+            }
+
+            // Add the contents for all of the sub-chapters within the
+            // current chapter.
+            for Chapter {
+                content,
+                source,
+                mut destination,
+            } in get_subcontent_for_chapter(chapter, directory, depth, 2)
+            {
+                let _: bool = destination.set_extension("pot");
+                let catalog = catalogs
+                    .entry(destination.clone())
+                    .or_insert_with(|| Catalog::new(generate_catalog_metadata(ctx)));
+
+                let source = ctx.config.book.src.join(&source);
+                for (lineno, extracted) in extract_messages(&content) {
+                    let msgid = extracted.message;
+                    let source = format!("{}:{}", source.display(), lineno);
+                    add_message(catalog, &strip_link(&msgid), &source, &extracted.comment);
+                }
             }
         }
     }
+    catalogs
+        .iter_mut()
+        .for_each(|(_key, catalog)| dedup_sources(catalog));
+    Ok(catalogs)
+}
 
-    dedup_sources(&mut catalog);
+/// A view into the relevant template information held by
+/// `mdbook::book::Chapter` and a location to store the exported polib
+/// messages.
+struct Chapter {
+    /// The chapter's content.
+    content: String,
+    /// The file where the content is sourced.
+    source: path::PathBuf,
+    /// The output destination for the polib template.
+    destination: path::PathBuf,
+}
 
-    Ok(catalog)
+// A recursive function to crawl a chapter's sub-items and get the
+// relevant info to produce a set of po template files.
+fn get_subcontent_for_chapter(
+    c: &book::Chapter,
+    provided_file_path: path::PathBuf,
+    provided_depth: usize,
+    depth: usize,
+) -> Vec<Chapter> {
+    if c.sub_items.is_empty() {
+        return Vec::new();
+    };
+
+    // Iterate through sub-chapters and pull the chapter content,
+    // path, and destination to store the template.
+    c.sub_items
+        .iter()
+        .filter_map(|item| {
+            let BookItem::Chapter(chapter) = item else {
+                return None;
+            };
+            let (chapter_info, new_path) = match &chapter.path {
+                Some(chapter_path) => {
+                    // Append the chapter's name to the template's
+                    // destination when the depth has not surpassed
+                    // the provided value.
+                    let destination = if depth < provided_depth {
+                        provided_file_path.join(slug(&chapter.name))
+                    } else {
+                        provided_file_path.clone()
+                    };
+
+                    let info = Chapter {
+                        content: chapter.content.clone(),
+                        source: chapter_path.clone(),
+                        destination: destination.clone(),
+                    };
+                    (Some(info), destination)
+                }
+                None => (None, provided_file_path.clone()),
+            };
+
+            // Recursively call to get sub-chapter contents.
+            Some(chapter_info.into_iter().chain(get_subcontent_for_chapter(
+                chapter,
+                new_path,
+                provided_depth,
+                depth + 1,
+            )))
+        })
+        .flatten()
+        .collect()
+}
+
+// Trim a string slice to only contain alphanumeric characters and
+// dashes.
+fn slug(title: &str) -> String {
+    // Specially handle "C++" to format it as "cpp" instead of "c".
+    let title = title.to_lowercase().replace("c++", "cpp");
+    title
+        .split_whitespace()
+        .map(|word| {
+            word.chars()
+                .filter(|&ch| ch == '-' || ch.is_ascii_alphanumeric())
+                .collect::<String>()
+        })
+        .filter(|word| !word.is_empty())
+        .collect::<Vec<_>>()
+        .join("-")
 }
 
 #[cfg(test)]
@@ -180,14 +345,29 @@ mod tests {
         std::fs::create_dir(tmpdir.path().join("src"))
             .context("Could not create src/ directory")?;
 
+        // Dynamically create the directories for the listed files.
+        // Write the contents to the file in the specified directory.
         for (path, contents) in files {
-            std::fs::write(tmpdir.path().join(path), contents)
-                .with_context(|| format!("Could not write {path}"))?;
+            let file_path = tmpdir.path().join(path);
+            let directory_path = file_path
+                .parent()
+                .context("File path unexpectedly ended in a root or prefix")?;
+
+            std::fs::create_dir_all(directory_path).context(format!(
+                "Could not create directory {}",
+                directory_path.display()
+            ))?;
+            std::fs::write(file_path.clone(), contents)
+                .with_context(|| format!("Could not write {}", file_path.display()))?;
         }
 
         let mdbook = MDBook::load(tmpdir.path()).context("Could not load book")?;
         let ctx = RenderContext::new(mdbook.root, mdbook.book, mdbook.config, "dest");
         Ok((ctx, tmpdir))
+    }
+
+    fn default_template_file() -> path::PathBuf {
+        path::PathBuf::from("messages.pot")
     }
 
     #[test]
@@ -241,7 +421,8 @@ mod tests {
         let (ctx, _tmp) =
             create_render_context(&[("book.toml", "[book]"), ("src/SUMMARY.md", "")])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string).unwrap();
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string).unwrap();
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(catalog.metadata.project_id_version, "");
         assert!(!catalog.metadata.pot_creation_date.is_empty());
         assert!(catalog.metadata.po_revision_date.is_empty());
@@ -264,7 +445,10 @@ mod tests {
             ("src/SUMMARY.md", ""),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string).unwrap();
+        // Using a depth of 0 to include all messages in a single
+        // template file.
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(catalog.metadata.project_id_version, "My Translatable Book");
         assert_eq!(catalog.metadata.language, "fr");
         Ok(())
@@ -299,7 +483,8 @@ mod tests {
             ("src/suffix.md", ""),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(
             catalog
                 .messages()
@@ -319,7 +504,7 @@ mod tests {
     }
 
     #[test]
-    fn test_create_catalog() -> anyhow::Result<()> {
+    fn test_create_catalogs() -> anyhow::Result<()> {
         let (ctx, _tmp) = create_render_context(&[
             ("book.toml", "[book]"),
             ("src/SUMMARY.md", "- [The *Foo* Chapter](foo.md)"),
@@ -332,7 +517,8 @@ mod tests {
             ),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
 
         for msg in catalog.messages() {
             assert!(!msg.is_translated());
@@ -366,7 +552,8 @@ mod tests {
             ),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(
             catalog
                 .messages()
@@ -401,7 +588,8 @@ mod tests {
             ),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(
             catalog
                 .messages()
@@ -439,7 +627,8 @@ mod tests {
             ),
         ])?;
 
-        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        let catalogs = create_catalogs(&ctx, std::fs::read_to_string)?;
+        let catalog = &catalogs[&default_template_file()];
         assert_eq!(
             catalog
                 .messages()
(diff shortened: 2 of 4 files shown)


antoniolinhart:
@mgeisler What do you think about the implementation of splitting the template file? In this version I chose to add messages that did not have a matching source to a catalog that was saved in the file name specified in the input params.

I'm open to changing how we use that pot-file input param, I thought that this made the most sense for now. Would you want to add another optional field for output directory if we have a depth > 0?

src/bin/mdbook-xgettext.rs
```diff
@@ -109,12 +191,49 @@ fn main() -> anyhow::Result<()> {
...
         .ok_or_else(|| anyhow!("Expected a string for output.xgettext.pot-file"))?;
+    // The depth on which to split the output file. 0 (default) is to include
+    // all messages into a single POT file, 1 will include each top-level file
+    // in a separate POT file, greater numbers will further split the main POT
+    // file.
+    let split_depth = cfg
+        .get("depth")
+        .unwrap_or(&Value::Integer(0))
+        .as_integer()
+        .ok_or_else(|| anyhow!("Expected a number for output.xgettext.depth"))?;
```
mgeisler, line 202: A small trick which allows us to avoid depending directly on the `toml` crate:

```suggestion
    let split_depth = cfg
        .get("depth")
        .and_then(|value| value.as_integer())
        .ok_or_else(|| anyhow!("Expected a number for output.xgettext.depth"))?;
```

It turns out that mdbook depends on a slightly old version of the toml crate (see https://github.com/rust-lang/mdBook/issues/2019 for why), so it feels ever-so-slightly better if we can avoid mentioning this version in our own `Cargo.toml` file.

src/bin/mdbook-xgettext.rs
```diff
@@ -98,6 +104,82 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+    }
+
+    // Source fields in the catalog messages are represented
+    // in a format like `src/{path to file}:{line num}`.
+    // Isolate the path to file and then add the message
+    // to the proper sub-catalog.
+    for message in catalog.messages() {
+        let source = message.source().to_owned();
+
+        let src_path = "src/";
```
mgeisler, line 155: I don't think we can assume that the path starts with `src/` — it's [configurable](https://rust-lang.github.io/mdBook/format/configuration/general.html#general-metadata).

I think you can get around with this by letting the `RenderContext` drive the process: instead of splitting a catalog after creating it, you create multiple catalogs, one per sub-tree of the book.

mgeisler:
> * Mimics the directory structure for the book

I think this is a key decision. I was thinking we would use the [`Book`](https://docs.rs/mdbook/latest/mdbook/book/struct.Book.html) (from `RenderContext`) as the driver of this — not the file names.

We currently iterate over the `BookItem`s in depth-first order (via [`Book::iter`](https://docs.rs/mdbook/latest/mdbook/book/struct.Book.html#method.iter)), but we could instead iterate over the `sections` field. I think that would let you avoid having to think of where `SUMMARY.md` fits into the picture.

As for what to name the POT files, I guess this can be derived from `Chapter::path`? Perhaps we need a top-level output directory (`po/`) and then we create `po/foo.pot` for a `Chapter { path: `foo.md` }` chapter. We would put all `sub_items` into the same `po/foo.pot` file if we're splitting at `depth = 1`. If we use `depth = 2`, then we would create `po/foo.pot` with only the data from this chapter and create `po/foo/bar.pot` with everything from the `sub_items`.

There is no guarantee that nested chapters are stored in nested files, but I think we should be safe if we follow the logical structure of the `Book` in this manner.

antoniolinhart:
> > * Mimics the directory structure for the book
> 
> I think this is a key decision. I was thinking we would use the [`Book`](https://docs.rs/mdbook/latest/mdbook/book/struct.Book.html) (from `RenderContext`) as the driver of this — not the file names.
> 
> We currently iterate over the `BookItem`s in depth-first order (via [`Book::iter`](https://docs.rs/mdbook/latest/mdbook/book/struct.Book.html#method.iter)), but we could instead iterate over the `sections` field. I think that would let you avoid having to think of where `SUMMARY.md` fits into the picture.

Oh! That's a good point for using the sections directly instead of that built in iterator. Until looking into the Book items struct it was hard to know exactly why that worked the way it did.

If I follow through all of the Chapters into their sub items and take the content from each Chapter and so on, that should include all the messages, right? I'm running into an interesting situation where there's *some* messages that aren't there that should be. Comparing between a current messages.pot and a messages.pot from my program with depth 0, it's missing about 10 messages that all appear to be the Rust scripts that get included as sample code. There are many scripts included in the messages that are sample code, there just happen to be some that don't come through?

> 
> As for what to name the POT files, I guess this can be derived from `Chapter::path`? Perhaps we need a top-level output directory (`po/`) and then we create `po/foo.pot` for a `Chapter { path: `foo.md` }` chapter. We would put all `sub_items` into the same `po/foo.pot` file if we're splitting at `depth = 1`. If we use `depth = 2`, then we would create `po/foo.pot` with only the data from this chapter and create `po/foo/bar.pot` with everything from the `sub_items`.
> 
> There is no guarantee that nested chapters are stored in nested files, but I think we should be safe if we follow the logical structure of the `Book` in this manner.

This makes more sense to follow the book instead. I think it will lead to the split depth input being a lot more logical too. Thank you!

Still working through the revised implementation!

mgeisler:
> There are many scripts included in the messages that are sample code, there just happen to be some that don't come through?

Much of the Rust code is included via the `{{#include foo.rs}}` system. That syntax is a [mdbook extension]https://rust-lang.github.io/mdBook/format/mdbook.html#including-files). This is done by the built-in [`links` preprocessor](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html).

Could the scripts be related to this?

Oh... another idea: are you perhaps seeing the effects of #75 by @dyoo? It was merged just a few days ago :slightly_smiling_face:

mgeisler:
> > There is no guarantee that nested chapters are stored in nested files, but I think we should be safe if we follow the logical structure of the `Book` in this manner.
> 
> This makes more sense to follow the book instead. I think it will lead to the split depth input being a lot more logical too. Thank you!
> 
> Still working through the revised implementation!

Sounds great, thank you so much for your help here!

mgeisler:
Hi @antoniolinhart, just a small thing: the format check fails. You should configure your editor to call `rustfmt` automatically (or run `dprint fmt` on the code).

antoniolinhart:
> > There are many scripts included in the messages that are sample code, there just happen to be some that don't come through?
> 
> Much of the Rust code is included via the `{{#include foo.rs}}` system. That syntax is a [mdbook extension]https://rust-lang.github.io/mdBook/format/mdbook.html#including-files). This is done by the built-in [`links` preprocessor](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html).
> 
> Could the scripts be related to this?
> 
> Oh... another idea: are you perhaps seeing the effects of #75 by @dyoo? It was merged just a few days ago 🙂

Looks like I need to pull `main` more often :laughing:  This seems to be what my issue was. Thank you!

src/bin/mdbook-xgettext.rs
```diff
@@ -17,16 +17,16 @@
 //! This program works like `xgettext`, meaning it will extract
 //! translatable strings from your book. The strings are saved in a
 //! GNU Gettext `messages.pot` file in your build directory (typically
-//! `po/messages.pot`).
+//! `po/messages.pot`). When the `depth` parameter is included, a new
+//! directory will contain the template files split based on the tiers
+//! of directory nesting.
```
mgeisler, line 22: I guess "directory" here is outdated now? :slightly_smiling_face:

src/bin/mdbook-xgettext.rs
```diff
@@ -17,16 +17,16 @@
...
 
 use anyhow::{anyhow, Context};
 use mdbook::renderer::RenderContext;
-use mdbook::BookItem;
+use mdbook::{book::Chapter, BookItem};
```
mgeisler, line 26: More of a FYI: If VS Code doesn't format this as

```suggestion
use mdbook::BookItem;
use mdbook::book::Chapter;
```

then I think it means that you are not using `rustfmt` from the nightly Rust channel. Try `rustup default nightly` and see if the formatting changes. The `rustfmt.toml` file is controlling this and we use a currently unstable option (which non-nightly `rustfmt`s will ignore with a warning).

The `dprint check` check in CI will use nightly Rust and should flag this as well (after you resolve the merge conflict — just rebase locally and throw away your old `Cargo.lock` file, it'll be regenerated when you run `cargo build`).

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+            let file_path = if depth == 0 {
+                Some(default_file_path.display().to_string())
+            } else if depth == 1 {
+                Some(chapter.path.clone().unwrap().display().to_string())
+            } else {
+                None
+            };
+            get_subcontent_for_chapter(chapter.clone(), file_path, depth)
+                .iter()
+                .for_each(
```
mgeisler, line 140: Please try converting this to a regular `for` loop — using `for_each` is almost never a nice syntax in my opinion. The extra indentation here is a sign that too much stuff is happening at once :slightly_smiling_face:

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                    },
+                );
         }
     }
 
-    Ok(catalog)
+    Ok(catalogs)
+}
+
+// Contains information necessary to create polib messages out
```
mgeisler, line 182: You're invited to make this a docstring with `///` — even though the type isn't public, it is simpler to just make all documentation comments docstrings.

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+    // The file where the content is sourced.
+    source: String,
+    // The output destination for the messages
+    // created with the content and source.
+    template_destination: String,
+}
+
+// Get ChapterTemplateInfo for sub-chapters, depending on the specified depth.
+fn get_subcontent_for_chapter(
+    c: Chapter,
```
mgeisler, line 197: From the name, I would expect `get_subcontent_for_chapter` to be a read-only operation on a `Chapter`. So this should most likely be `chapter: &Chapter` to show that we only borrow the chapter in a non-destructive sense. This would immediately allow you to avoid deep-cloning the chapter at the call site above.

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+    return extract_chapter_template_info(c, file_path, depth, 1);
+}
+
+// A recursive function to crawl a chapter's sub-items and get
+// the relevant info to produce a set of po template files.
+fn extract_chapter_template_info(
+    c: Chapter,
+    file_path: Option<String>,
+    depth: i64,
+    current_depth: i64,
```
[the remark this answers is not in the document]
mgeisler: ... or supply `1`, if that's the correct starting value :smile:

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                        Some(rel_path.display().to_string())
+                    } else {
+                        None
+                    };
+
+                    file_path2
+                }
+                None => file_path.clone(),
+            };
+            vec.append(&mut extract_chapter_template_info(
```
mgeisler, line 240: An `append` inside a `for_each` can normally be turned into a `map`. Here you have an `if` statement at the top, so you should use [`filter_map`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.filter_map) to express this more directly: you _filter_ because you ignore some items and you _map_ because you transform them into `extract_chapter_template_info(...)`.

src/bin/mdbook-xgettext.rs
```diff
@@ -81,21 +90,242 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                current_depth + 1,
+            ));
+        }
+    });
+    vec
+}
+
+// Split a main catalog into sub-catalogs, using a provided split
+// depth to separate catalogs based on directory nesting.
+fn _split_catalog_on_depth(
```
mgeisler, line 253: Identifiers starting with `_` are normally ignored in Rust — I don't know if this is still needed?

src/bin/mdbook-xgettext.rs
```diff
@@ -109,12 +339,34 @@ fn main() -> anyhow::Result<()> {
...
-    polib::po_file::write(&catalog, &output_path)
-        .with_context(|| format!("Writing messages to {}", output_path.display()))?;
+    let catalogs = create_catalogs(&ctx, split_depth, path).context("Extracting messages")?;
+
+    for (file_rel_path, catalog) in catalogs {
+        let mut file_path = PathBuf::from(file_rel_path);
+        // Template files should have the "pot" extension.
+        assert!(file_path.set_extension("pot"));
+
+        let directory_path = {
```
mgeisler, line 358: I think this is what [`parent()`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.parent) does?

src/bin/mdbook-xgettext.rs
```diff
@@ -124,30 +376,49 @@ mod tests {
...
+    }
+
     #[test]
     fn test_create_catalog_defaults() -> anyhow::Result<()> {
         let (ctx, _tmp) =
             create_render_context(&[("book.toml", "[book]"), ("src/SUMMARY.md", "")])?;
 
-        let catalog = create_catalog(&ctx).unwrap();
+        let catalogs = create_catalogs(&ctx, 0, "messages.pot")?;
+        let catalog = catalogs.get("messages.pot").unwrap();
```
mgeisler, line 421: I believe `map.get(key).unwrap()` is the same as `map[key]`.

src/bin/mdbook-xgettext.rs
```diff
@@ -222,4 +497,172 @@ mod tests {
...
+                ("src/foo/bar.md:3", "The first paragraph about Bar."),
+                ("src/foo/bar/baz.md:1", "How to Baz"),
+                ("src/foo/bar/baz.md:3", "The first paragraph about Baz."),
+            ]
+        );
+
+        Ok(())
+    }
+
+    // Helper constants written as functions.
```
mgeisler, line 560: I'm not entirely sure what this is doing — I have a strong feeling that we don't need to care about `const` here?

src/bin/mdbook-xgettext.rs
```diff
@@ -222,4 +497,172 @@ mod tests {
...
+    ]; "depth_3")]
+    #[test_case(4, vec![
+        (summary_file(), summary_content()),
+        (foo_file(), foo_content()),
+        (bar_file(), bar_content()),
+        (baz_file(), baz_content())
+    ]; "depth_greater_than_necessary")]
+    fn test_split_catalog_nested_directories(
+        split_depth: i64,
+        expected_message_tuples: Vec<(String, Vec<String>)>,
```
mgeisler, line 623: If the problem is that `&str` doesn't work here instead of `String`, then perhaps `&'static str` works?

In general, `&str` is a borrowed string slice: someone has allocated string data and a `&str` refers to this. There is an ellided lifetime here, so the type is actually `&'a str` for some lifetime `'a` (often a generic parameter to the function where the type is a parameter).

When you use a string literal, the Rust compiler is the one allocating the string data (somewhere in the text segment of your binary). All string literals have the type `&'static str`, with a special `'static` lifetime, signifying that they are valid for the entire program duration and never go out of scope.

mgeisler:
Hey @antoniolinhart, I skimmed through the updated code — from the test cases it looks like it works? Super cool!

You should rebase onto the latest `main` and resolve the merge conflict.

src/bin/mdbook-xgettext.rs
antoniolinhart: I was able to re-work the expected values to use the `&'static str` - I had trouble trying to see if I could do the same with mapping the Catalog messages to msgids. Do you know if it's possible? The compiler tells me that the Catalog does not live long enough (looking at the call to msgids_from_catalog, where I turn that fn into returning a `Vec<&'static str>`

src/bin/mdbook-xgettext.rs
antoniolinhart: Fixed!

src/bin/mdbook-xgettext.rs
antoniolinhart: Removed this comment - my reason for including it was to explain why the 'helper' variables were extracted into functions since they can't directly be written as variables

.vscode/settings.json
mgeisler, line 1: Could you instead add `.vscode/` to your [global Git ignore file]?

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -83,21 +104,148 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
             };
+
+            let rel_file_path = if depth == 0 {
+                default_file_path.to_path_buf()
+            } else {
+                catalogs.insert(path.clone(), Catalog::new(generate_catalog_metadata(&ctx)));
+                path.clone()
+            };
+
+            let catalog: &mut Catalog = catalogs.get_mut(&rel_file_path).with_context(|| {
```
mgeisler, line 133: Just use `catalogs[&rel_file_path]` here: it is a program error if this fails, not a user error. So it's fine to panic and crash the program.

In general, you want to use `Result` for error handling when the errors are _expected_. This includes things like a file not existing or a TCP connection being dropped. Not being able to find something in a hashmap (which you just inserted!) is not expected — it's a bug in the program and the user cannot fix it by running the program again with different input.

i18n-helpers/Cargo.toml
```diff
@@ -19,6 +19,7 @@ pulldown-cmark-to-cmark = "11.0.0"
 regex = "1.9.4"
 semver = "1.0.16"
 serde_json = "1.0.91"
+test-case = "3.1.0"
```
mgeisler, line 22: This is a test dependency, right? Then it should go into the `dev-dependencies` below.

However, I don't think we need extra complex machinery like this for our tests. Keeping things simple is normally better (the discussion about `&'static str` makes me feel that the new code isn't simple).

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -111,12 +259,34 @@ fn main() -> anyhow::Result<()> {
         .ok_or_else(|| anyhow!("Missing output.xgettext.pot-file config value"))?
         .as_str()
         .ok_or_else(|| anyhow!("Expected a string for output.xgettext.pot-file"))?;
+    // The depth on which to split the output file. 0 (default) is to include
```
mgeisler, line 262: Zero is not actually the default value, I get an error if I run this without a `depth` field :slightly_smiling_face: 

You will need something a bit more complex, something along the lines of

```rust
let depth = match cfg.get("depth") {
  None => 0,
  Some(depth) => depth.as_integer().with_context("Expected a number for output.xgettext.depth"))?,
};
```

mgeisler:
Hi @antoniolinhart, thanks for the cleanup. I tried checking out the branch and ran

```
% MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot", "depth": 1}}' mdbook build -d po1 
2023-09-29 10:38:28 [INFO] (mdbook::book): Book building has started
2023-09-29 10:38:28 [INFO] (mdbook::book): Running the xgettext backend
2023-09-29 10:38:28 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
```

This gives me

```
% tree po1                                                                  (~/src/comprehensive-rust)
po1
└── src
    ├── android
    │   ├── aidl.pot
    │   ├── build-rules.pot
    │   ├── interoperability.pot
    │   ├── logging.pot
    │   └── setup.pot
    ├── android.pot
```

I had hoped there would be a single POT file for each chapter in the top-level. Like this:

```
% tree expected1                                                            (~/src/comprehensive-rust)
expected1
├── android.pot
├── async.pot
├── bare-metal.pot
├── basic-syntax.pot
├── cargo.pot
```

I suppose the problem in an off-by-one error somewhere :slightly_smiling_face:

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -226,4 +422,168 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_create_catalog_nested_directories() -> anyhow::Result<()> {
+        let (ctx, _tmp) = create_render_context(&[
```
mgeisler, line 428: If you don't need `_tmp`, then just use `_` as a wildcard pattern to throw away the binding.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -226,4 +422,168 @@ mod tests {
...
+            assert!(message_tuples.contains(&message_tuple));
+        }
+
+        Ok(())
+    }
+
+    fn summary_file() -> PathBuf {
+        PathBuf::from("src/SUMMARY.md")
+    }
+    fn summary_content() -> Vec<&'static str> {
```
mgeisler, line 491: Please drop this — it makes the tests hard to understand since it now requires knowledge of the `test_case` attribute below. I would prefer having two simple tests over this :slightly_smiling_face:

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -111,12 +259,34 @@ fn main() -> anyhow::Result<()> {
         .ok_or_else(|| anyhow!("Missing output.xgettext.pot-file config value"))?
```
mgeisler, line 259: I think we could make things more uniform by removing the `pot-file` config value — replace it by `messages.pot`.

People should then instead use the `-d` or `--dest-dir` argument to `mdbook build` to specify where the POT files should go. So

```
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' mdbook build -d po
```
today becomes

```
MDBOOK_OUTPUT='{"xgettext": {"depth": 0}}' mdbook build -d po
```

(or rather `MDBOOK_OUTPUT='{"xgettext": {}}'` when you make `depth` default to zero).

src/bin/mdbook-xgettext.rs
mgeisler: A large part of learning a new language is to figure out all the idioms and shortcuts :smile:

src/bin/mdbook-xgettext.rs
mgeisler: I'll take a look!

src/bin/mdbook-xgettext.rs
mgeisler: My advice would be to not use `test_case`. Just create a very simple unit test with static strings. Avoid abstracting too much in test — use string literals everywhere to make the test very clear.

Basically, apply the opposite of the normal DRY principle :smile:

antoniolinhart:
> Hi @antoniolinhart, thanks for the cleanup. I tried checking out the branch and ran
> 
> ```
> % MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot", "depth": 1}}' mdbook build -d po1 
> 2023-09-29 10:38:28 [INFO] (mdbook::book): Book building has started
> 2023-09-29 10:38:28 [INFO] (mdbook::book): Running the xgettext backend
> 2023-09-29 10:38:28 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
> ```
> 
> This gives me
> 
> ```
> % tree po1                                                                  (~/src/comprehensive-rust)
> po1
> └── src
>     ├── android
>     │   ├── aidl.pot
>     │   ├── build-rules.pot
>     │   ├── interoperability.pot
>     │   ├── logging.pot
>     │   └── setup.pot
>     ├── android.pot
> ```
> 
> I had hoped there would be a single POT file for each chapter in the top-level. Like this:
> 
> ```
> % tree expected1                                                            (~/src/comprehensive-rust)
> expected1
> ├── android.pot
> ├── async.pot
> ├── bare-metal.pot
> ├── basic-syntax.pot
> ├── cargo.pot
> ```
> 
> I suppose the problem in an off-by-one error somewhere 🙂

The challenge with this is that Android isn't actually a top-level chapter, mdbook treats it as a part title and doesn't contain the sub-items as expected like "interoperability" "aidl", "build-rules". I think if we wanted something like this we'd have to leverage the listed directory in the code, or change the SUMMARY file to make it easier

mgeisler:
> The challenge with this is that Android isn't actually a top-level chapter, mdbook treats it as a part title and doesn't contain the sub-items as expected like "interoperability" "aidl", "build-rules". I think if we wanted something like this we'd have to leverage the listed directory in the code, or change the SUMMARY file to make it easier

Oh! I had not realized this until now...

We can change our own `SUMMARY.md` file, but we won't be able to the change the [upstream format](https://rust-lang.github.io/mdBook/format/summary.html) easily.

I think the code right now splits the POT file into ~63 files, whereas we only have ~13 parts. Do you think you can add code for tracking the parts? I would suggest transforming the part title into a slug using something like this:

```rust
fn slug(title: &str) -> String {
    title
        .split_whitespace()
        .map(|word| {
            word.to_lowercase()
                .chars()
                .filter(|&ch| ch == '-' || !ch.is_ascii_punctuation())
                .collect::<String>()
        })
        .filter(|word| !word.is_empty())
        .collect::<Vec<_>>()
        .join("-")
}
```

I tested it a bit on our own titles plus some headings from the Rust Book and they look reasonable.

antoniolinhart:
> Hi @antoniolinhart, thanks for the cleanup. I tried checking out the branch and ran
> 
> ```
> % MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot", "depth": 1}}' mdbook build -d po1 
> 2023-09-29 10:38:28 [INFO] (mdbook::book): Book building has started
> 2023-09-29 10:38:28 [INFO] (mdbook::book): Running the xgettext backend
> 2023-09-29 10:38:28 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
> ```
> 
> This gives me
> 
> ```
> % tree po1                                                                  (~/src/comprehensive-rust)
> po1
> └── src
>     ├── android
>     │   ├── aidl.pot
>     │   ├── build-rules.pot
>     │   ├── interoperability.pot
>     │   ├── logging.pot
>     │   └── setup.pot
>     ├── android.pot
> ```
> 
> I had hoped there would be a single POT file for each chapter in the top-level. Like this:
> 
> ```
> % tree expected1                                                            (~/src/comprehensive-rust)
> expected1
> ├── android.pot
> ├── async.pot
> ├── bare-metal.pot
> ├── basic-syntax.pot
> ├── cargo.pot
> ```
> 
> I suppose the problem in an off-by-one error somewhere 🙂

After the most recent improvements I think it's closer to expectations - depth 1 splits each 'part' into a separate catalog.

Here's the output I get for the following depths (using the Android chapter as an example after depth = 1):

# Depth 1 (note that summary.pot includes messages from any chapters prior to encountering a PartTitle):

```
$ MDBOOK_OUTPUT='{"xgettext": {"depth": 1}}'   mdbook build -d po/XX
2023-10-05 17:36:46 [INFO] (mdbook::book): Book building has started
2023-10-05 17:36:46 [INFO] (mdbook::book): Running the xgettext backend
2023-10-05 17:36:46 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
```

```
$ tree po/XX
po/XX
├── android.pot
├── bare-metal-afternoon.pot
├── bare-metal-morning.pot
├── concurrency-afternoon.pot
├── concurrency-morning.pot
├── day-1-afternoon.pot
├── day-1-morning.pot
├── day-2-afternoon.pot
├── day-2-morning.pot
├── day-3-afternoon.pot
├── day-3-morning.pot
├── final-words.pot
├── solutions.pot
└── summary.pot
```

# Depth 2

```
$ MDBOOK_OUTPUT='{"xgettext": {"depth": 2}}'   mdbook build -d po/XX
2023-10-05 17:36:46 [INFO] (mdbook::book): Book building has started
2023-10-05 17:36:46 [INFO] (mdbook::book): Running the xgettext backend
2023-10-05 17:36:46 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
```

```
$ tree po/XX                                                                   
po/XX                                                                           
├── android                                                                     
│   ├── aidl.pot                                                                
│   ├── build-rules.pot                                                         
│   ├── exercises.pot                                                           
│   ├── interoperability.pot                                                    
│   ├── logging.pot                                                             
│   ├── setup.pot                                                               
│   └── welcome.pot
```

# Depth 3

```
$ MDBOOK_OUTPUT='{"xgettext": {"depth": 3}}'   mdbook build -d po/XX
2023-10-05 17:36:46 [INFO] (mdbook::book): Book building has started
2023-10-05 17:36:46 [INFO] (mdbook::book): Running the xgettext backend
2023-10-05 17:36:46 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
```

```
$ tree po/XX                                                                   
po/XX                                                                           
├── android                                                                     
│   ├── aidl                                                                    
│   │   ├── changing-api.pot                                                    
│   │   ├── client.pot                                                          
│   │   ├── deploy.pot                                                          
│   │   ├── implementation.pot                                                  
│   │   ├── interface.pot                                                       
│   │   └── server.pot                                                          
│   ├── aidl.pot                                                                
│   ├── build-rules                                                             
│   │   ├── binary.pot                                                          
│   │   └── library.pot                                                         
│   ├── build-rules.pot                                                         
│   ├── exercises.pot                                                           
│   ├── interoperability                                                        
│   │   ├── with-c.pot                                                          
│   │   └── with-java.pot                                                       
│   ├── interoperability.pot                                                    
│   ├── logging.pot                                                             
│   ├── setup.pot                                                               
│   └── welcome.pot
```

# Depth 4

```
$ MDBOOK_OUTPUT='{"xgettext": {"depth": 4}}'   mdbook build -d po/XX
2023-10-05 17:36:46 [INFO] (mdbook::book): Book building has started
2023-10-05 17:36:46 [INFO] (mdbook::book): Running the xgettext backend
2023-10-05 17:36:46 [INFO] (mdbook::renderer): Invoking the "xgettext" renderer
```

```
$ tree po/XX                                                                   
po/XX                                                                           
├── android                                                                     
│   ├── aidl                                                                    
│   │   ├── changing-api.pot                                                    
│   │   ├── client.pot                                                          
│   │   ├── deploy.pot                                                          
│   │   ├── implementation.pot                                                  
│   │   ├── interface.pot                                                       
│   │   └── server.pot                                                          
│   ├── aidl.pot                                                                
│   ├── build-rules                                                             
│   │   ├── binary.pot                                                          
│   │   └── library.pot                                                         
│   ├── build-rules.pot                                                         
│   ├── exercises.pot                                                           
│   ├── interoperability                                                        
│   │   ├── with-c                                                              
│   │   │   ├── calling-c-with-bindgen.pot                                      
│   │   │   └── calling-rust-from-c.pot                                         
│   │   ├── with-c.pot                                                          
│   │   └── with-java.pot                                                       
│   ├── interoperability.pot                                                    
│   ├── logging.pot                                                             
│   ├── setup.pot                                                               
│   └── welcome.pot
```

src/bin/mdbook-xgettext.rs
antoniolinhart: Split out into multiple tests using the same "book"!

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Fixed this in the most recent version - thanks for noting this. I agree it makes more sense without! I've found the directory argument makes most sense now with a `po/XX` (at least in the usage for comprehensive-rust) so all the newly created files can be separated from the language files in the main `po` directory

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -56,6 +61,18 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
     metadata.mime_version = String::from("1.0");
     metadata.content_type = String::from("text/plain; charset=UTF-8");
     metadata.content_transfer_encoding = String::from("8bit");
+    metadata
+}
+
+/// Transform a Book into a HashMap of (file_name, catalog) pairs.
+/// The depth will be used to crawl the chapters and sub-chapters
```
mgeisler, line 68: Tiny thing: the docstrings are Markdown and start with a one-line description of the function. You can see the results of this with `cargo doc --open` locally:

```suggestion
/// Transform a Book into a HashMap of `(file_name, catalog)` pairs.
///
/// The depth will be used to crawl the chapters and sub-chapters
```

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
         add_message(&mut catalog, line, &source);
     }
 
+    let mut catalogs = HashMap::new();
+    // The catalog from the summary data will exist in the single pot file for
+    // a depth of 0, will exist in a top-level separate `summary.pot` file for
+    // a depth of 1, or exist within in a `summary.pot` file within the default
+    // directory for chapters without a corresponding part title.
+    let mut current_top_level = "summary".to_owned();
+    let summary_template_destination = if depth == 0 {
```
mgeisler, line 112: This is perfectly fine, but in case you didn't know, you could use a `match` here:

```rust
match depth {
  0 => ...
  1 => ...
  _ => ...
}
```

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                template_destination,
+            } in get_subcontent_for_chapter(chapter, template_destination, depth, 2)
+            {
+                if let None = catalogs.get(&template_destination) {
+                    catalogs.insert(
+                        template_destination.clone(),
+                        Catalog::new(generate_catalog_metadata(&ctx)),
+                    );
+                }
+                let catalog = catalogs.get_mut(&template_destination).unwrap();
```
mgeisler, line 171: This is a great example of a place where the [entry API](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#examples) can be used. You should be able to use `catalogs.entry(template_destination).or_insert_with(|| ...)` instead.

Two advantages of this:

- You avoid hashing the key twice
- You avoid the `unwrap()` call at the end — unwraps are always scary since they are places where the program can crash.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                let source = ctx.config.book.src.join(&source);
+
+                for (lineno, msgid) in extract_messages(&content) {
+                    let source_with_lineno = format!("{}:{}", source.display(), lineno);
```
mgeisler, line 175: Would you be okay to use `path` instead of `source` and then rename your `source_with_lineno` back to `source`? That would align with the [`source()` method in polib](https://docs.rs/polib/latest/polib/message/trait.MessageView.html#tymethod.source) (that's where I got my variable names from in the first place).

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
         }
     }
 
-    Ok(catalog)
+    Ok(catalogs)
+}
+
+/// A view into the relevant template information held by `mdbook::book::Chapter`
+/// and a location to store the exported polib messages.
+struct ChapterTemplateInfo {
```
mgeisler, line 187: The word "template" here and elsewhere confuses me — possibly because I've done a lot of web development. For me, a "template" is something you later instantiate and fill out with concrete values.

So a "chapter template" is not clear to me since the chapters are already full of content?

Could we call the struct `Chapter` instead since I think it's a minimal version of `BookItem::Chapter` enum variant? The `template_destination` field could then conveniently become `destination`.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+                        };
+
+                        let info = ChapterTemplateInfo {
+                            content: chapter.content.clone(),
+                            source: chapter_path.clone(),
+                            template_destination: destination.clone(),
+                        };
+                        (Some(info), destination)
+                    }
+                    None => (None, provided_file_path.clone()),
```
mgeisler, line 231: I think you don't need to clone here: it looks like the last usage of `provided_file_path` and you have full ownership of it. So you can pass along the ownership directly (if I read things correctly).

Try running `cargo clippy` if you haven't already done so.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -86,21 +103,161 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
+
+                // Recursively call to get sub-chapter contents.
+                Some(
+                    get_subcontent_for_chapter(&chapter, new_path, provided_depth, depth + 1)
+                        .into_iter()
+                        .chain(chapter_info),
+                )
+            } else {
+                None
+            }
```
mgeisler, line 242: Could this be turned into an early return? The [`let ... else` syntax](https://google.github.io/comprehensive-rust/control-flow/if-let-expressions.html) supports this in newer Rust versions (see the speaker notes there).

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -109,17 +266,31 @@ fn main() -> anyhow::Result<()> {
         .config
         .get_renderer("xgettext")
         .ok_or_else(|| anyhow!("Could not read output.xgettext configuration"))?;
-    let path = cfg
-        .get("pot-file")
-        .ok_or_else(|| anyhow!("Missing output.xgettext.pot-file config value"))?
-        .as_str()
-        .ok_or_else(|| anyhow!("Expected a string for output.xgettext.pot-file"))?;
```
mgeisler, line 116: I love getting rid of this, but please update the command line here

https://github.com/google/mdbook-i18n-helpers/blob/1480904ad45f2dab4949d30826335a0a7a3b96fb/i18n-helpers/USAGE.md?plain=1#L79-L82

as well.

It would be good to also mention something about the `depth` parameter there.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -109,17 +266,31 @@ fn main() -> anyhow::Result<()> {
...
-    let output_path = ctx.destination.join(path);
-    let catalog = create_catalog(&ctx).context("Extracting messages")?;
-    polib::po_file::write(&catalog, &output_path)
-        .with_context(|| format!("Writing messages to {}", output_path.display()))?;
+    let catalogs = create_catalogs(&ctx, depth).context("Extracting messages")?;
+
+    // Create a template file for each entry with the content from the respective catalog.
+    for (mut file_path, catalog) in catalogs {
+        // Template files should have the "pot" extension.
+        assert!(file_path.set_extension("pot"));
```
mgeisler, line 285: Two small things:

- Asserts are normally only used in tests so I would not assert anything here.
- I would suggest letting the code that forms `file_path` in `create_catalogs` do this. That way the code would take full responsibility for forming a useful path.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -229,4 +425,573 @@ mod tests {
...
+                "src/foo.md",
+                "# How to Foo\n\
+                 \n\
+                 The first paragraph about Foo.\n",
+            ),
+            (
+                "src/foo/bar.md",
+                "# How to Bar\n\
+                 \n\
+                 The first paragraph about Bar.\n",
```
mgeisler, line 449: Since you're not testing the text extraction functionality as such, I would suggest making the files here have just a single line.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -229,4 +425,573 @@ mod tests {
...
+        ];
+
+        let message_tuples = catalog
+            .messages()
+            .map(|msg| (msg.source(), msg.msgid()))
+            .collect::<Vec<(&str, &str)>>();
+
+        for message_tuple in expected_message_tuples {
+            assert!(message_tuples.contains(&message_tuple));
+        }
```
mgeisler, line 486: I think you can assert something stronger: you know the order in which you walk the tree of chapters, so you know the order of the messages in `catalog`.

The order is important since it directly affects the lines written into the POT files. So it is important that it is not accidentally randomized via iteration over the hashmap.

Basically, I suggest

```rust
assert_eq!(
    catalog.messages().map(|msg| msg.source()).collect::<Vec<_>>(),
    vec![
        "src/SUMMARY.md:1",
        ...
    ]
)
```

While writing this, I figured that we might not need to look at the `msgid` fields at all — there is probably another test which checks them?

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -229,4 +425,573 @@ mod tests {
...
+                vec![
+                    "Intro",
+                    "Foo",
+                    "The Foo Chapter",
+                    "The Bar Section",
+                    "The Baz Subsection",
+                    "Foo Exercises",
+                    "Intro to X",
+                    "In this course, you will learn X.",
+                ],
```
mgeisler, line 556: Oh nice, this is much closer to what I suggested above :smile: 

I think it could be made a tiny bit clearer if you tested against the sources.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -229,4 +425,573 @@ mod tests {
...
+                PathBuf::from("foo/foo-exercises"),
+                vec!["Exercises on Foo", "Try out this Foo example."],
+            ),
+        ]);
+
+        // The message ids in a catalog do not have a constant order,
+        // so check for the presence of each one in the vector.
+        assert_eq!(expected_message_tuples.keys().len(), message_tuples.len());
+        for (file_path, msgids) in message_tuples {
+            let expected_msgids = expected_message_tuples.get(&file_path).unwrap();
```
mgeisler, line 781: Doing `map.get("key").unwrap()` is the same as `map["key"]` so I would prefer the shorter version.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -229,4 +425,573 @@ mod tests {
...
+                ],
+            ),
+            (
+                PathBuf::from("foo/foo-exercises"),
+                vec!["Exercises on Foo", "Try out this Foo example."],
+            ),
+        ]);
+
+        // The message ids in a catalog do not have a constant order,
+        // so check for the presence of each one in the vector.
```
mgeisler, line 778: This looks like a crucial detail! I would have expected the order to be deterministic (following the order in which you traverse sub-chapters)?

mgeisler (changes requested):
Hi @antoniolinhart, thanks for all the work put into this! I think the functionality is super now, I just have a bunch of small comments about naming and a few Rust tricks you can apply.

I hope you can incorporate it over the next few weeks.

mgeisler:
Hi @antoniolinhart! Happy new year :partying_face:

Would you be able to rebase this PR on top of the latest `main`? We should focus on getting it merged to prevent it from bit-rotting!

Even if the functionality to merge the split PO files are not yet there, the splitting functionality could be useful to someone (they would then have to write a tiny script to `msgcat` everything back together before using the translation).

antoniolinhart:
> Hi @antoniolinhart! Happy new year 🥳
> 
> Would you be able to rebase this PR on top of the latest `main`? We should focus on getting it merged to prevent it from bit-rotting!
> 
> Even if the functionality to merge the split PO files are not yet there, the splitting functionality could be useful to someone (they would then have to write a tiny script to `msgcat` everything back together before using the translation).

Happy new year! I got taken away with some high priority team projects so it's been a while since I've been able to look at this. That's a good point! My goal will be to (this week/next week) get the comments responded to / get it rebased on `main` since I know the longer I wait the harder it will be :sweat_smile: 

Thanks for your patience on this!

mgeisler:
> Happy new year! I got taken away with some high priority team projects so it's been a while since I've been able to look at this.

That's completely okay! The functionality here is luckily not blocking anything right now.

> My goal will be to (this week/next week) get the comments responded to / get it rebased on `main` since I know the longer I wait the harder it will be 😅

Thanks for sticking with it! I think the conflicts should be small at this point, so it should be possible to revive this PR without too much trouble.

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Trying to remember my thinking from 4 month younger me~

I'm not sure what I meant, as this doesn't appear to be the case currently. We do need to iterate through the catalogs in the Map and match them with the file that they represent, though

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Done

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: I re-worked this to test against the source string instead like you suggested above

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Good point about enforcing that messages are in order. And yes there is another test that deals with the `msgid` fields, so we don't need that here

Done!

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: I won't oppose making the tests a bit shorter :smile:

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Fair point about not using asserts in the executable, I had it there since it feels a bit better than dropping the boolean. But no problem to remove it

And I moved the `set_extension` to the `create_catalogs` code to have a better separation of the caller and the function body

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: No problem at all! I can definitely see how template would mean something completely different in that context. `Chapter` is more clear here

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: No problem, done!

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Thanks! I know about the entry API but still have a lot to learn about the special functions and their uses. This shortens this code block by a bunch

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: I like it! Added

i18n-helpers/src/bin/mdbook-xgettext.rs
antoniolinhart: Ooh~ I was not aware that this was how that was generated How do I rebuild this doc to reflect local changes? It seems to be for ToT and not based on my local branch, even when I `cargo build`

antoniolinhart:
Should be ready for another round of review @mgeisler ! Thank you for your extended patience and support!

mgeisler:
> Should be ready for another round of review @mgeisler ! Thank you for your extended patience and support!

I'm the one thanking! Let me look it through now.

i18n-helpers/src/xgettext.rs
```diff
@@ -98,6 +91,23 @@ where
     metadata.mime_version = String::from("1.0");
     metadata.content_type = String::from("text/plain; charset=UTF-8");
     metadata.content_transfer_encoding = String::from("8bit");
+    metadata
+}
+
+/// Build catalog from RenderContext
+///
+/// # Arguments
+
```
mgeisler, line 100:
```suggestion
```

mgeisler (approved):
Thank you so much for this, let's merge this now and then see about the combination side of things later!

i18n-helpers/src/xgettext.rs
```diff
@@ -398,4 +574,435 @@ mod tests {
...
+        ];
+
+        let message_tuples = catalog
+            .messages()
+            .map(|msg| (msg.source(), msg.msgid()))
+            .collect::<Vec<(&str, &str)>>();
+
+        for message_tuple in expected_message_tuples {
+            assert!(message_tuples.contains(&message_tuple));
+        }
```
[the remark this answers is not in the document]
antoniolinhart: Thanks for your suggestion here! This was masking a hidden issue of my breadth-first-search adding the messages in reverse breadth-first-search order :smiley:

i18n-helpers/src/xgettext.rs
antoniolinhart: Done

antoniolinhart:
@mgeisler 

Should be good to go! It appears that the diff base is not against the correct source so it's showing that I changed a _lot_ of files that I obtained from rebasing on the updated code. I tried looking around but it appears that this merge is into `google:main` which should have the updated files?

mgeisler:
> @mgeisler
> 
> Should be good to go!

Great!

> It appears that the diff base is not against the correct source so it's showing that I changed a _lot_ of files that I obtained from rebasing on the updated code. I tried looking around but it appears that this merge is into `google:main` which should have the updated files?

Yeah, something is off here — I'll try to untangle it and then merge the PR. It'll take me a day of two to get around to it :smile:

antoniolinhart:
> > @mgeisler
> > Should be good to go!
> 
> Great!
> 
> > It appears that the diff base is not against the correct source so it's showing that I changed a _lot_ of files that I obtained from rebasing on the updated code. I tried looking around but it appears that this merge is into `google:main` which should have the updated files?
> 
> Yeah, something is off here — I'll try to untangle it and then merge the PR. It'll take me a day of two to get around to it 😄

Thanks!! Also the clippy failure was due to me leaving some debugging logs in by accident (after the last time I ran clippy) so that should be fixed now :+1:

antoniolinhart:
> > Thanks!! Also the clippy failure was due to me leaving some debugging logs in by accident (after the last time I ran clippy) so that should be fixed now 👍
> 
> Hey Antonio, I think I got your changes back into a simple branch. I don't actually know if you did anything wrong here. What I did on my side was to
> 
> 1. Create a temporary branch at the point where your `split-pot-file` branch was.
> 2. Hard reset `split-pot-file` to `main` to throw away all commits and merges and what not.
> 3. Soft reset `split-pot-file` back to the temporary branch. This added all the changes back, but as a delta to `main`.
> 4. Commit this under your name and force push back to your repository.
> 
> If the tests pass, then I'll merge this next.

Wonderful! Thank you I will try this out next time I run into the same issue.

Issue #13, opened 2023-04-04
Add tool which can build translations for all available languages


mgeisler:
The `mdbook-gettext` preprocessor translates a book into a single language. To translate your book into all available languages, you need to build a look yourself. An example of this can be found in the [`publish.yml` GitHub action for Comprehensive Rust 🦀](https://github.com/google/comprehensive-rust/blob/907608c504d3dfd45d1569d7968f13c709c7d917/.github/workflows/publish.yml#L43-L51):

```yaml
      - name: Build all translations
        run: |
          for po_lang in ${{ env.LANGUAGES }}; do
              echo "::group::Building $po_lang translation"
              MDBOOK_BOOK__LANGUAGE=$po_lang \
              MDBOOK_OUTPUT__HTML__SITE_URL=/comprehensive-rust/$po_lang/ \
              mdbook build -d book/$po_lang
              echo "::endgroup::"
          done
```

We should make this easier somehow. Idea:

- Build a small command line tool (perhaps called `mdbook-i18n`) where `mdbook-i18n build` would do the looping seen above.

mgeisler:
> Maybe related with upstream `mdbook`? [rust-lang/mdBook#2029](https://github.com/rust-lang/mdBook/pull/2029)

Yes, partially: @jooyunghan also works in Android and made the PR to simplify our translation infrastructure.

Here, I was thinking of making a small stand-alone tool which can do all the looping we need to implement translations. It would

- call `MDBOOK_BOOK__LANGUAGE=xx mdbook build` for all languages.
- it would generate the [language picker drop-down](https://github.com/google/comprehensive-rust/blob/5cdb73387fba83d625db39d80c82688a0070f31a/theme/index.hbs#L156-L166) and inject this into the generated HTML (or theme or similar).

I think that approach would be more flexible than trying to teach mdbook all this.

mgeisler:
Note that I've made implementing this issue a bit more complicated with https://github.com/google/comprehensive-rust/pull/1243: you'll need to use Git library to update the `src/` folder to the right version.

On the other hand, this feature makes the backend here much more useful: instead of replacing a small for-loop, it can now replace more complex logic! This makes this renderer more relevant and useful for everybody.

Issue #35, opened 2023-02-15
Add links between translations


mgeisler:
We need to add links between the translations. This will take two forms:

1. A language picker for users. This is done via google/comprehensive-rust#411.
2. Links for robots so that search engines can index the pages reliably.

The second part is done by adding

```html
<link rel="alternate" hreflang="en" href="https://google.github.io/comprehensive-rust/" />
<link rel="alternate" hreflang="pt-BR" href="https://google.github.io/comprehensive-rust/pt-BR/" />
```

to the `head` element of every page. The alternatives will link each page with its translated siblings. See [this documentation](https://developers.google.com/search/docs/specialty/international/localized-versions) for details.

mgeisler:
One way of solving this would be to write a tool which can add the necessary HTML after the book has been rendered. The advantages of this would be

- It would give us an extensible way to post-process translations without having to change the upstream `mdbook` every time.
- It would give us an entry point for the user. I'm thinking that `mdbook-i18n build` would build all translations, `mdbook-i18n serve` would show them all in your browser, etc... This would give us fine-grained control over things. This would also be how we can implement #16.

sakex:
The first thing we should do is add a way to embed a JSON config inside of the page. Then we'll be able to use the data from Javascript to add the links.

My initial idea was to convert `book.toml` to a JSON and add that JSON as a script tag inside of the HTML rendered template. This would not be ideal because we would potentially write unwanted data to the HTML. The fix is to write a subset of `book.toml`.

The [Config struct](https://github.com/rust-lang/mdBook/blob/master/src/config.rs#L69) has a `rest` member that contains the remaining data inside of `book.toml`. We could add a `[js_data]` field to `book.toml` that we could read from our custom renderer to add extra data to the page.

sakex:
I'm starting to work on this. My first thought is that we could move the translation logic to the renderer so we wouldn't need to do this as part of the CI with a different command. WDYT?

mgeisler:
> My first thought is that we could move the translation logic to the renderer so we wouldn't need to do this as part of the CI with a different command. WDYT?

Using a preprocessor to do translations works really well and gives us a nice and important property: the search index is local to the current language (as one would expect). Similar for the `print.html` page that is generated. We can also have different assets for different languages.

The architecture of mdbook is that you work with _one_ book — so I would not suggest changing this right now.

Also, as you know (#70), we don't yet have very good control over the renderer :smile: So I'm not sure how this would actually work?

[...]

PR #80, opened 2023-09-15
Create mdbook-tera-backend renderer

### LINKED ISSUE #70: Write `mdbook` renderer with strong templating system
Currently, the `mdbook` templates are very limited: https://github.com/rust-lang/mdBook/issues/2016. I believe this is on purpose to prevent templates from becoming too unwieldy.

However, a limited template engine only works when it's easy to modify the logic behind the templates — which means modifying the upstream `mdbook` code. There are a number of features which we would like to implement, but which require modifying the generated HTML:

- #35
- #12

The common theme in these issues is that they require us to inject more data into the template (a list of languages, a canonical URL) and that we need to loop over these values and extract various parts.

A stronger [`mdbook` renderer](https://rust-lang.github.io/mdBook/format/configuration/renderers.html) would allow us to do this inside the theme itself. The renderer should be generic and expose the same values to the theme as the `HtmlHandlebars` renderer does today — plus data configured in the `book.toml` file. This would be how users will inject more data.

The templating engine should then allow users to

- define helper functions (Handlebars comes with a [limited set of built-in helpers](https://docs.rs/handlebars/latest/handlebars/#built-in-helpers)).
  - A much-needed function would be a function which computes a consistent canonical URL for a given `.md` file path, e.g., it would turn `/index.html` to just `/`. This is the [`url` function in Django](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#url).
- read and write variables (Handlebars seems to only work with the variables defined in the context passed in from `mdbook`).
- include extra files defined in the theme (Handlebars seems to only work with an existing set of files defined in the Rust code).

Ideally, the new renderer would be very small: it converts the chapter content from Markdown to HTML and sends this plus the chapter information to the template engine. All of the [static file copying](https://github.com/rust-lang/mdBook/blob/02f3823e4caae28cc0b1ceac284f81743f8fd29b/src/renderer/html_handlebars/hbs_renderer.rs#L230-L292) in `HtmlHandlebars` would be replaced by the theme.

## Template Engine

Take a look at https://lib.rs/template-engine and https://crates.io/categories/template-engine. I would look at [askama](https://lib.rs/crates/askama) and [tera](https://lib.rs/crates/tera), both of which are inspired by Django. Since we target offline-rendering, we don't need the fastest engine in the world — we need something which is flexible and extensible without having to touch the `mdbook` renderer all the time.

Ideally, we would prototype a set of template helper functions in a theme first (in Comprehensive Rust, most likely). If they become widely useful, we will move those helper definitions to a "standard library" for the new renderer. This way others can reuse the helpers in their own themes.

sakex:
Here is the prototype, it does almost everything we want already, I just need to port the actual LanguagePicker

You can see that the code to create a custom component is pretty straightforward (see the test component in src/custom_components/test_component.rs).

You can check the test in src/custom_component_renderer/book_directory_renderer.rs to see how it all works.

Fixes #70.


src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+                custom_component.render(dom_manipulator, &self.config)?;
+            }
+        }
+        self.file_buffer_cache.clear();
+        self.file_buffer_cache = document.html();
+        Ok(())
+    }
+
+    fn process_file(&mut self, path: &Path) -> Result<()> {
+        if path.extension().unwrap_or_default() != "html" {
```
mgeisler, line 60: Perhaps this works:
```suggestion
        if path.extension() != Some("html") {
```

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            let mut file = fs::File::open(path)?;
+            file.read_to_string(&mut self.file_buffer_cache)?;
+        }
+        self.render_components()?;
+        let mut file = fs::OpenOptions::new()
+            .write(true)
+            .truncate(true)
+            .open(path)?;
+        file.write_all(self.file_buffer_cache.as_bytes())?;
```
mgeisler, line 73: I have seen that mdbook has some utilities for writing files. I would like to use them if possible — I think they log something at a debug log level.

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+        if path.extension().unwrap_or_default() != "html" {
+            return Ok(());
+        }
+        self.file_buffer_cache.clear();
+        {
+            let mut file = fs::File::open(path)?;
+            file.read_to_string(&mut self.file_buffer_cache)?;
+        }
+        self.render_components()?;
+        let mut file = fs::OpenOptions::new()
```
mgeisler, line 69: These open options are common, can you use `std::fs::create` instead?

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+                node_ids.push(node.id());
+            }
+            let tree = &mut document.tree;
+            for id in node_ids {
+                let dom_manipulator = NodeManipulator::new(tree, id);
+                custom_component.render(dom_manipulator, &self.config)?;
+            }
+        }
+        self.file_buffer_cache.clear();
+        self.file_buffer_cache = document.html();
```
mgeisler, line 55: You lose the cache right here because you assign a new string to it.

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    fn process_file(&mut self, path: &Path) -> Result<()> {
+        if path.extension().unwrap_or_default() != "html" {
+            return Ok(());
+        }
+        self.file_buffer_cache.clear();
+        {
+            let mut file = fs::File::open(path)?;
+            file.read_to_string(&mut self.file_buffer_cache)?;
```
mgeisler, line 66: The read here is used later in `render_components` below? The connection needs to be made explicit, otherwise it can break in a small refactoring.

I would remove the cache and only add it later if you can show a speedup to justify it.

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn test_render_book() {
+        use super::*;
+        use crate::custom_components::test_component::TestComponent;
+        use std::fs::File;
+        use std::io::Write;
+        use tempfile::tempdir;
+
+        let dir = tempdir().unwrap();
+        {
+            let mut file = File::create(dir.path().join("test.html")).unwrap();
```
mgeisler, line 105: There is a `std::fs::write` helper for writing a string to a file.

src/custom_component_renderer/dom_manipulator.rs
```diff
@@ -0,0 +1,147 @@
+use super::error::RendererError;
+use crate::custom_component_renderer::error::Result;
+use ego_tree::{NodeId, NodeMut, NodeRef, Tree};
+use markup5ever::namespace_url;
+use markup5ever::{ns, Attribute, LocalName, QualName};
+use scraper::node::{Element, Text};
+use scraper::Node;
+
+pub struct NodeManipulator<'a> {
```
mgeisler, line 9: When I saw this struct used in the other file, I had expected it to come from a library. Why are we writing all this code? 🙂

src/custom_component_renderer/languages_configuration.rs
mgeisler, line 1: This file is way too small to carry it's own weight. The whole thing can be made much more compact — like the code I sketches in the bug using Tera.

src/custom_components/test_component.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub struct TestComponent {}
+
+impl TestComponent {
+    pub fn new() -> TestComponent {
+        TestComponent {}
+    }
+}
+
+impl Component for TestComponent {
```
mgeisler, line 19: Does this show that people need to write Rust code to introduce new components?

If so, that is problematic. The goal was for us/people to push logic into the templates and to avoid being constrained by the hard-coded functionality offered to the templates.

mgeisler:
Hey @sakex, thanks for putting this up! I think my only major concern is how people (myself included! 🙂) will create and modify the custom components?

I had hoped they would be defined in the theme, just like I see Tera allows you to [create new macros](https://keats.github.io/tera/docs/#macros)  in a template file.

I feel this is crucial since I need a way of looking at the output when modify the templates. So I need to be able to run `mdbool serve` and see the output right away.

sakex:
A separate custom-made component is definitely doable.

It could look like

```
book.toml

[output.i18-helper.components]
MyCustomComponent = "path/to/comp.html"
```

```
comp.html

<div>
some stuff in there, maybe tera templates too
</div>
```

```
index.html
...
<MyCustomComponent />
...
```

sakex:
Because in my test component, I'm generating the HTML using DOM manipulations with the scraper crate. But we don't have to do it that way, we can also do HTML::parse on a string and then inject the parsed node to the original tree.

sakex:
For complex and common components, we will create them inside of the crate. Otherwise, users will be able to create their own. I'll add that feature in a further CL though, if that works for you. 

LanguagePicker should definitely live here for instance. Also I'd really like to create an <ExercizeCodeBlock /> and a  <WasmCodeBlock /> in the future. Because the components created here don't need to be used in the template, they can also be added to the markdown directly

mgeisler:
> A separate custom-made component is definitely doable.
> 
> It could look like
> 
> ```
> book.toml
> 
> [output.i18-helper.components]
> MyCustomComponent = "path/to/comp.html"
> ```
> 
> ```
> comp.html
> 
> <div>
> some stuff in there, maybe tera templates too
> </div>
> ```
 
This is the main question: how do I create a component with custom _logic_? With loops, variables, if statements, etc? I have the impression that I need to write Rust code for that?

I liked using Tera much more: there we can keep _our_ code to a minimum. I don't really like code: more code is normally more trouble :smile: Put differently, I don't want us to maintain a template system here. Pushing all logic to the template library should make our life easier.

It also makes things clearer for downstream users: just modify `index.hbs` to hook in the call that loads a macro from `mdbook-i18n-helpers.html` and then use those macros where you like. I think it would even be possible to let downstream users override parts of the macros using the inheritance system in Tera, but I'm not 100% sure.

sakex:
Let's do it that way then, I think it will look pretty goodOn Sep 16, 2023 16:54, Martin Geisler ***@***.***> wrote:

A separate custom-made component is definitely doable.
It could look like
book.toml

[output.i18-helper.components]
MyCustomComponent = "path/to/comp.html"

comp.html

<div>
some stuff in there, maybe tera templates too
</div>


This is the main question: how do I create a component with custom logic? With loops, variables, if statements, etc? I have the impression that I need to write Rust code for that?
I liked using Tera much more: there we can keep our code to a minimum. I don't really like code: more code is normally more trouble 😄 Put differently, I don't want us to maintain a template system here. Pushing all logic to the template library should make our life easier.
It also makes things clearer for downstream users: just modify index.hbs to hook in the call that loads a macro from mdbook-i18n-helpers.html and then use those macros where you like. I think it would even be possible to let downstream users override parts of the macros using the inheritance system in Tera, but I'm not 100% sure.

src/custom_component_renderer/book_directory_renderer.rs
sakex: Types are not compatible

src/main.rs
```diff
@@ -0,0 +1,49 @@
...
+        ctx.config.book.language.clone(),
+        root,
+    );
+    let default_language = renderer_config.i18n.default_language.clone();
+
+    let mut renderer = BookDirectoryRenderer::new(renderer_config);
+    let components = create_components();
+
+    for component in components {
+        renderer.add_component(component);
```
sakex, line 42: This will be used to add custom component

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,264 @@
...
+
+            Command::new("mdbook")
+                .arg("build")
+                .arg(&book_folder)
+                .arg("-d")
+                .arg(&format!("{destination}/{identifier}"))
+                .env("MDBOOK_BOOK__LANGUAGE", identifier)
+                .env(
+                    "MDBOOK_OUTPUT__HTML__SITE_URL",
+                    &format!("/comprehensive-rust/{identifier}/"),
```
mgeisler, line 128: This looks hard-coded to work with Comprehensive Rust :slightly_smiling_face: The tooling here is used by 5+ other projects right now (see the README), so it has to be generic.

sakex: Yes, this part is not ready yet. I will remove the need to spawn new processes by directly calling `book.clone()` -> `get_text(book)` ->  `book.build()` but I first need to move the methods in `get_text` out into their own library

sakex: Also now, with the `mdbook build` process we don't get hot reloading because it's not added by default

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,264 @@
...
+            if let Some(default_language) = default_language {
+                if default_language == identifier {
+                    continue;
+                }
+            }
+
+            let destination = self.config.destination.as_path().display().to_string();
+            let book_folder = self.config.root.as_path().display().to_string();
+
+            Command::new("mdbook")
```
mgeisler, line 120: This should not call out to a `mdbook` binary — what if people run `mdbook` as `cargo run` from a `mdbook` clone? If we need to invoke `mdbook`, we should instead invoke the functionality using the library functions in `mdbook`, similar to how `mdbook-i18n` does it.

However, I'm not sure we need to invoke `mdbook` in the first version: it would already be super helpful if we keep the renderer as a per-language renderer.

The simple for-loop outside of `mdbook build` is not so bad, and #13 is about making a wrapper tool which can call `mdbook build` for all languages. This tool could also handle `mdbook serve` then (probably by building all languages and starting a simple webserver).

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,264 @@
...
+                .arg("-d")
+                .arg(&format!("{destination}/{identifier}"))
+                .env("MDBOOK_BOOK__LANGUAGE", identifier)
+                .env(
+                    "MDBOOK_OUTPUT__HTML__SITE_URL",
+                    &format!("/comprehensive-rust/{identifier}/"),
+                )
+                .output()?;
+
+            std::fs::rename(
```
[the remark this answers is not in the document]
mgeisler: I'm not super sure how the reload functionality works...

But I'm sure we should not hard-code the structure used by Comprehensive Rust. Other projects will likely use `en.example.net` and `fr.example.net` sub-domains and other projects will make other choices: lower-case language codes (`zh-cn` instead of `zh-CN`), write the language in full, etc.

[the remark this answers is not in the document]
sakex: I know how it works, I just need to add a member variable to the book as they do in `mdbook::serve`

src/custom_component_renderer/book_directory_renderer.rs
sakex: > This should not call out to a mdbook binary — what if people run mdbook as cargo run from a mdbook clone? If we need to invoke mdbook, we should instead invoke the functionality using the library functions in mdbook, similar to how mdbook-i18n does it.

It's really how `mdbook` itself does it, so I think it's fine. But I will anyway replace this child process with simple library calls, which will give more control, performance and stability. I will remove the need to spawn new processes by directly calling `book.clone()` -> `get_text(book)` -> `book.build()` but I first need to move the methods in `get_text` out into their own library

> However, I'm not sure we need to invoke mdbook in the first version: it would already be super helpful if we keep the renderer as a per-language renderer.

I can take it out of the first version. We'll see how it pans out in the end, my goal is to support both per language and global languages builds, so that it will be nicer for translators to debug directly from `mdbook serve`.

> The simple for-loop outside of mdbook build is not so bad, and https://github.com/google/mdbook-i18n-helpers/issues/13 is about making a wrapper tool which can call mdbook build for all languages. This tool could also handle mdbook serve then (probably by building all languages and starting a simple webserver).

My next commit will fix https://github.com/google/mdbook-i18n-helpers/issues/13 as well. The added benefits, beyond having one less step in github workflows, is that all the configurations will live in one place (`book.toml`), and `mdbook serve` will actually serve all the translations. Currently, if we want to add a new language, we need to update the workflow as well as `index.hbs`.  With this change, all would be handled for us and we would only need to care about the actual translation work.

## Next step

This week (probably Friday, unless I get some time before), I will replace the `Command::new` call with library calls. I will also add the `custom components` feature. Then we can see how we split the PR and decide  what features we don't want. My main concern is how complex the code will become, I think it will be ~800 LOC without tests and `Cargo.lock` (next commit will be around 80 LOC + moving ~100 LOC from `bin/get_text.rs` to `lib.rs`.)

src/custom_component_renderer/book_directory_renderer.rs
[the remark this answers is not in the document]
mgeisler: Sounds great!

src/lib.rs
```diff
@@ -339,7 +339,7 @@ fn is_nontranslatable_codeblock_group(events: &[(usize, Event)]) -> bool {
             // Heuristic to check whether the codeblock nether has a
             // literal string nor a line comment.  We may actually
             // want to use a lexer here to make this more robust.
-            !codeblock_text.contains("\"") && !codeblock_text.contains("//")
+            !codeblock_text.contains('"') && !codeblock_text.contains("//")
```
mgeisler, line 342: Looks like a good, but unrelated change — please submit it in a separate PR and have @dyoo approve it :smile:

src/custom_component_renderer/book_directory_renderer.rs
```diff
@@ -0,0 +1,280 @@
...
+        self.book.config.book.language = original_language;
+        self.book.config.build.build_dir = book_dir.to_owned();
+        Ok(())
+    }
+
+    pub(crate) fn render_book(&mut self) -> Result<()> {
+        let html_dir = self.book_dir.join("html");
+        if !html_dir.is_dir() {
+            return Err(RendererError::InvalidPath(format!(
+                "{:?} is not a directory",
```
mgeisler, line 125: Are you assuming here that people first run `mdbook build` with the normal `html` renderer?

Should you not call the renderer yourself so that you don't need to error out here?

src/lib.rs
mgeisler: Thanks for #85.

sakex:
@mgeisler I have implemented all the features we wanted (load template, dependencies, attributes from HTML). 

You can see an example in action there:

https://github.com/google/comprehensive-rust/pull/1216

sakex:
We are missing documentation,  I will add it after we've finalized the conversation. I will also create a book inside of this repository for testing purposes in a future CL

mdbook-tera-backend/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mdbook-tera-backend"
+version = "0.0.1"
+authors = ["Martin Geisler <mgeisler@google.com>"]
```
mgeisler, line 4: You should add your own name here :smile:

mdbook-tera-backend/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+categories = ["template-engine"]
+edition = "2021"
+keywords = ["mdbook", "tera", "renderer", "template"]
+license = "Apache-2.0"
+repository = "https://github.com/google/mdbook-i18n-helpers"
+description = "Plugins for a mdbook translation workflow based on Gettext."
```
mgeisler, line 10: The description should be made specific to the new backend.

mgeisler:
Hey @sakex, thanks for putting this up! I think it is simpler than before, but I still think it can become even simpler.

> We are missing documentation, I will add it after we've finalized the conversation. I will also create a book inside of this repository for testing purposes in a future CL

I see many new files introduced here with lots of Rust code. This raises questions for me:

- How will _other projects_ use this? Remember that the new backend is a binary for them, so they cannot extend it with Rust code. They cannot implement the trait you define so it seems they cannot create new custom components?
- Why do we need a custom component?
  - Is Tera not strong enough to let me implement the [language picker](https://github.com/google/comprehensive-rust/blob/main/theme/index.hbs#L170-L189) and #35 using only a Tera template?
  - If it does not support this, then we should not use Tera. Those two issues are the high-priority goals of writing a new rendering engine. They are the only reason we care to write code for this :slightly_smiling_face:

In my mind, the "shape" of a new renderer should be so that it

- only depends on `tera` or any other template engine (no HTML parsing needed).
- doesn't know anything about i18n.
- lets users (including us) build up a vocabulary of commonly used functions in the template language (not Rust). I imagine this simply means that the user tells you about a directory where the template macros live. You then add that to an include path and the user can then define everything in there.

When I see more code than this, I conclude that the renderer does more than it has to do.

As an example, I would expect the work in #84 to build on top of such a bare-bones renderer by supplying it with template macros that know about i18n. The backend that runs `mdbook build` for every language _must_ know about i18n, but the `mdbook-tera-backend` must not. That is really the test which checks if the design is loosely coupled and if the components are reusable.

sakex:
> Hey @sakex, thanks for putting this up! I think it is simpler than before, but I still think it can become even simpler.
> 
> > We are missing documentation, I will add it after we've finalized the conversation. I will also create a book inside of this repository for testing purposes in a future CL
> 
> I see many new files introduced here with lots of Rust code. This raises questions for me:
> 
> * How will _other projects_ use this? Remember that the new backend is a binary for them, so they cannot extend it with Rust code. They cannot implement the trait you define so it seems they cannot create new custom components?

There is no trait anymore, everything is done via templates. Please take a look at https://github.com/google/comprehensive-rust/pull/1216

> * Why do we need a custom component?

It's a struct not a trait, just an internal that let people define their new components

>   * Is Tera not strong enough to let me implement the [language picker](https://github.com/google/comprehensive-rust/blob/main/theme/index.hbs#L170-L189) and [Add links between translations #35](https://github.com/google/mdbook-i18n-helpers/issues/35) using only a Tera template?

It is, please take a look at https://github.com/google/comprehensive-rust/pull/1216

>   * If it does not support this, then we should not use Tera. Those two issues are the high-priority goals of writing a new rendering engine. They are the only reason we care to write code for this 🙂
> 
> In my mind, the "shape" of a new renderer should be so that it
> 
> * only depends on `tera` or any other template engine (no HTML parsing needed).

The HTML part is a very low amount of code, but it makes the template system much cleaner, wrapping tera in handlebars just looks very bad. I have tried, and I don't think that many people will want to put up with it.

> * doesn't know anything about i18n.

It doesn't

> * lets users (including us) build up a vocabulary of commonly used functions in the template language (not Rust). I imagine this simply means that the user tells you about a directory where the template macros live. You then add that to an include path and the user can then define everything in there.

Look at https://github.com/google/comprehensive-rust/pull/1216, there are macros and includes supported

> When I see more code than this, I conclude that the renderer does more than it has to do.
> 
> As an example, I would expect the work in #84 to build on top of such a bare-bones renderer by supplying it with template macros that know about i18n. The backend that runs `mdbook build` for every language _must_ know about i18n, but the `mdbook-tera-backend` must not. That is really the test which checks if the design is loosely coupled and if the components are reusable.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,133 @@
...
+            .map(|s| tera::Value::String(s.to_owned()))
+            .ok_or_else(|| tera::Error::from(format!("Could not strip prefix")))
+    }
+}
+
+pub struct CustomComponent {
+    template: Tera,
+    name: String,
+    /// Used to generate unique ids for each component to prevent collisions in javascript with query selectors.
+    counter: RefCell<u32>,
```
mgeisler, line 37: This code looks very custom to how you will use it in Comprehensive Rust?

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::collections::{BTreeMap, HashMap};
+use std::path::{Path, PathBuf};
+
+use tera::Tera;
+
+use crate::{RendererError, Result};
+
+use super::RenderingContext;
+
+fn make_strip_prefix_function() -> impl tera::Function {
```
mgeisler, line 12: You're "cheating" here by defining a helper — other users cannot do this, so we should avoid doing it too.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    fn create_context(
+        &self,
+        rendering_context: &RenderingContext,
+        attributes: BTreeMap<String, String>,
+    ) -> tera::Context {
+        let counter = self.counter.replace_with(|&mut counter| counter + 1);
+        let mut context = tera::Context::new();
+        context.insert("counter", &counter);
+        context.insert("language", &rendering_context.language);
```
mgeisler, line 68: This is about i18n, right? :slightly_smiling_face: I'm expecting to see the renderer here dump the entire `book.toml` into the context — a very hands-off approach which allows you to make the renderer completely generic.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: They have a built in `trim_start_matches` (https://keats.github.io/tera/docs/) so we can remove this, we probably don't need extra methods

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: Yes it's extracted from the book, so we can also remove

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub path: PathBuf,
+
+    #[serde(default)]
+    pub dependencies: Vec<String>,
+}
+
+#[derive(Deserialize)]
+pub struct TeraRendererConfig {
+    pub components: Vec<TeraComponentConfig>,
+}
```
mgeisler, line 107: This is the kind of infrastructure that I don't expect there. In my mind, the renderer should just iterate over the HTML files and process the one by one. It will probably be a 20-30 line renderer at the end :slightly_smiling_face:

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: I think it's essential if you're adding a script with `getElementById` and you want to add the same component multiple times in the page, they will all have a different id thanks to `{{counter}}` so it's definitely something that people will want

mdbook-tera-backend/src/tera_renderer/error.rs
```diff
@@ -0,0 +1,25 @@
...
+    HtmlRewritingError(#[from] lol_html::errors::RewritingError),
+    #[error("Mdbook Error: {0}")]
+    Mdbook(#[from] mdbook::errors::Error),
+    #[error("Error in strip_prefix call: {0}")]
+    StripPrefixError(#[from] StripPrefixError),
+    #[error("Serde error: {0}")]
+    SerdeError(#[from] serde_json::Error),
+    #[error("Serde error: {0}")]
+    DependencyNotFound(String),
+}
```
mgeisler, line 23: Similar to above, we should not have to maintain this kind of code. Every line of code is a liability, so we should avoid writing more than we have to.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,158 @@
...
+                })
+            })
+            .collect();
+        let output = lol_html::rewrite_str(
+            file_content,
+            RewriteStrSettings {
+                element_content_handlers: custom_components_handlers,
+                ..RewriteStrSettings::default()
+            },
+        )?;
```
mgeisler, line 155: I bet using a `lol_html` library makes it hard to use this renderer to process other kinds of files (text files)?

If you instead first let the `html` renderer output to a directory, then it will copy over all other files too. Then you can iterate over all `.html`, `.txt`, `.js`, ... files and run them through the template engine.

sakex: It's a feature we can add in the future without much difficulty, just don't use HTML for non html files based on the extension. That could be useful for JS, but I'm not sure it's something we need

mdbook-tera-backend/src/tera_renderer/error.rs
sakex: We could use anyhow, but I'm not a fan

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: It's needed to specify the path to templates to include and their dependencies, it is required by Tera

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl CustomComponent {
+    pub fn new(name: &str, template_str: &str, dependencies: &[&Self]) -> Result<CustomComponent> {
+        let mut template = Tera::default();
+        for dep in dependencies {
+            template.extend(&dep.template)?;
+        }
+        template.add_raw_template(name, template_str)?;
+        template.register_function("strip_prefix", make_strip_prefix_function());
```
mgeisler, line 47: This is the thing which others cannot do when they use our tool.

sakex: This one is not needed because there is already a builtin for this. The other method is a nice helper for our specific use case. In the end, I don't think it is unreasonable that people will want us to add new methods. But tera already has a lot of builtins (even `now` and `date` for instance)

mdbook-tera-backend/src/tera_renderer/custom_component.rs
mgeisler: > same component

My concern is that you're inventing new vocabulary here — don't design "web components", just make a trivial system which runs Tera on the HMTL. The simplest thing which will let us solve the issues that motivate this effort.

If people want to call the same Tera macro multiple times, then they can set and maintain a counter, no?

mdbook-tera-backend/src/tera_renderer/custom_component.rs
mgeisler: > In the end, I don't think it is unreasonable that people will want us to add new methods. But tera already has a lot of builtins (even `now` and `date` for instance)

Yes and no... the problem for me is ownership and maintenance. If someone needs a `fetch_stock_quote` helper function, should we add it? We'll have to maintain it going forward and I don't really want to do that.

Delegating everything to the template system makes the decision easy: we surface relevant data and then people can manipulate it as much as they want using functions/macros in the template system.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
mgeisler: Are you saying that a Tera template cannot include files which haven't been explicitly configured in Tera somehow? Can we use an include directory instead? Just define that all files in `theme/templates/` are available for use in the templates. That should simplify the setup.

mdbook-tera-backend/src/tera_renderer/error.rs
mgeisler: We should just remove it, the code here should not be so complex that we need this much infrastructure.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: Maybe, I haven't had time to dive on all the features, so it might be possible to add a global variable like this but I don't know how

mdbook-tera-backend/src/tera_renderer/error.rs
sakex: Sounds good, let's just use anyhow

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: It works too, the tradeoff is that we'll need to read the directory and extract file names from there, and since all templates will have access to one another, they could have cyclical dependencies (which is handled by tera)

mdbook-tera-backend/src/tera_renderer/custom_component.rs
sakex: Yes, I agree and I can't think of a use case yet for a domain specific method

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,29 @@
+mod tera_renderer;
+
+use mdbook::renderer::RenderContext;
+use std::io;
+
+use crate::tera_renderer::*;
```
mgeisler, line 6: Please avoid glob imports — they make it hard to review the code in places like GitHub.

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+
+fn main() {
+    let mut stdin = io::stdin();
+    // Get the configs
```
mgeisler, line 10:
```suggestion
```

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,152 @@
...
+            .ctx
+            .destination
+            .parent()
+            .ok_or_else(|| {
+                anyhow!(
+                    "Destination directory {:?} has no parent",
+                    self.ctx.destination
+                )
+            })?
+            .to_owned();
```
mgeisler, line 72: Can we just `unwrap()` and also avoid the `to_owned()`? It looks like a lot of code for a simple operation.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn process_file(&mut self, path: &Path) -> Result<()> {
+        if path.extension().unwrap_or_default() != "html" {
+            return Ok(());
+        }
+        let mut file_content = String::new();
+        {
+            let mut file = fs::File::open(path)?;
+            file.read_to_string(&mut file_content)?;
+        }
```
mgeisler, line 100: I think we talked about this before? There is [`std::fs::read_to_string`](https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html) for this.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let output = self.render_components(&file_content, path)?;
```
mgeisler, line 102: Please rip out the notion of a "custom component" for now.

Instead, require people to write pure Tera code in the HTML files which you process. That would be the MVP for this feature. I would like us to land the core feature first before we solve potential down-stream problems.

mgeisler (changes requested):

mdbook-tera-backend/src/tera_renderer/renderer.rs
mgeisler: I feel we're reviewing the same lines again and again because conversations are resolved without actually being fixed?

mgeisler:
Hey @sakex, please add docstrings to all functions — explain why the function is there and how they fit into the bigger picture.

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,118 @@
...
+#[serde(untagged)]
+pub enum Component {
+    Named { name: String, path: PathBuf },
+    Anonymous(PathBuf),
+}
+
+/// Configuration in `book.toml` `[output.tera-renderer]`.
+#[derive(Deserialize)]
+pub struct TeraRendererConfig {
+    /// Relative path to the templates directory.
```
mgeisler, line 55: Relative path from where?

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,118 @@
...
+                        template.add_template_file(path, Some(name))?;
+                        CustomComponent::new(name, template)?
+                    }
+                    Component::Anonymous(path) => {
+                        let mut template = tera_template.clone();
+                        let name = path
+                            .file_stem()
+                            .unwrap_or_default()
+                            .to_str()
+                            .unwrap_or_default();
```
mgeisler, line 96: The `unwrap_or_default` looks a bit strange here: you'll end up with `""` if the filename contains non-UTF-8 bytes, right? That seems like an error situation which should just be handled with `?` instead?

mdbook-tera-backend/src/tera_renderer/renderer.rs
sakex: Yes it probably got reverted, I had some issues with Git the other day

mdbook-tera-backend/src/tera_renderer/renderer.rs
sakex: to_owned was needed because it borrows from self otherwise

src/custom_component_renderer/book_directory_renderer.rs
mgeisler: This is still an important question: we cannot rely on "chance" here :slightly_smiling_face:

sakex: Can't we simply define that as invalid and tell people they need to enable the HTML renderer? mdbook-pdf does that

mgeisler: My concern is the ordering: if the HTML renderer runs after this renderer, what kind of output do you get?

Let us imagine you run `mdbook build` $n$ times. I am afraid that the new renderer will use the HTML output from run $n-1$ when it produces output for run $n$. That will put us into the same annoying situation as when using LaTeX where you need to compile 2-3 times to be "really sure" that the output is stable.

sakex: No, it's guaranteed to run in that order with the current implementation because mdbook calls the renderers in alphabetical order. What's not ideal is that it's not documented, so it's Hyrum law for now. I will open a PR to document that behaviour when we submit this PR.

sakex: I'd rather even add a `after` parameter in their `book.toml` params

mgeisler: > I will open a PR to document that behaviour when we submit this PR.

It's great that you checked this — please open the PR already now.

We would also need to document this very carefully in our own code so that we know that the naming is important :grimacing:

mgeisler (approved):
Thanks for all the hard work! I'm excited to release this and start using it in Comprehensive Rust!

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,84 @@
...
+and configure the place where youre templates will live.
+For instance `theme/templates`:
+
+```toml
+...
+
+[output.tera-backend]
+templates_dir = "theme/templates"
+
+...
```
mgeisler, line 34: Perhaps this would be a good place to show that the `html` backend must be enabled:

```suggestion
[output.html] # You must still enable the html backend.

[output.tera-backend]
templates_dir = "theme/templates"
```

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,38 @@
+use anyhow::Result;
+use serde::Deserialize;
+use std::path::{Path, PathBuf};
+use tera::Tera;
+
+/// Configuration in `book.toml` `[output.tera-renderer]`.
+#[derive(Deserialize)]
+pub struct TeraRendererConfig {
+    /// Relative path to the templates directory from the `book.toml` directory.
+    pub templates_dir: PathBuf,
```
mgeisler, line 10: This should be `template_dir`, even though there are multiple templates in the directory (a bit like it's "shoe maker" and not "shoes maker" even though the person probably makes many shoes).

mdbook-tera-backend/src/tera_renderer/custom_component.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl TeraRendererConfig {
+    /// Recursively add all templates in the `templates_dir` to the `tera_template`.
+    fn add_templates_recursively(tera_template: &mut Tera, directory: &Path) -> Result<()> {
```
mgeisler, line 15: This should just be a regular function, no need to make it an associated function.

mdbook-tera-backend/src/tera_renderer/mod.rs
mgeisler, line 1: I've used the Rust 2018 style `foo.rs` and `foo/child.rs` naming scheme for all other files. Please rename this to `src/tera_renderer.rs` to follow this.

mdbook-tera-backend/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+categories = ["template-engine"]
+edition = "2021"
+keywords = ["mdbook", "tera", "renderer", "template"]
+license = "Apache-2.0"
+repository = "https://github.com/google/mdbook-i18n-helpers"
+description = "Plugin to extend mdbook with Tera templates and custom HTML components."
+
+[dependencies]
+anyhow = "1.0.75"
+chrono = { version = "0.4.31", default-features = false, features = ["alloc"] }
```
mgeisler, line 14: I don't think you need this dependency — please go through the others as well and remove everything not needed :slightly_smiling_face:

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,84 @@
+# [Tera](https://github.com/Keats/tera) backend extension for `mdbook`'s HTML backend
```
mgeisler, line 1: Please keep headings simple and to-the-point:

```suggestion
# Tera backend for `mdbook`
```

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+[![Visit crates.io](https://img.shields.io/crates/v/mdbook-i18n-helpers?style=flat-square)](https://crates.io/crates/mdbook-tera-backend)
+[![Build workflow](https://img.shields.io/github/actions/workflow/status/google/mdbook-i18n-helpers/test.yml?style=flat-square)](https://github.com/google/mdbook-i18n-helpers/actions/workflows/test.yml?query=branch%3Amain)
+[![GitHub contributors](https://img.shields.io/github/contributors/google/mdbook-i18n-helpers?style=flat-square)](https://github.com/google/mdbook-i18n-helpers/graphs/contributors)
+[![GitHub stars](https://img.shields.io/github/stars/google/mdbook-i18n-helpers?style=flat-square)](https://github.com/google/mdbook-i18n-helpers/stargazers)
+
+This `mdbook` backend makes it possible to use [tera](https://github.com/Keats/tera)
+templates and expand the capabilities of your books. It works on top of the default HTML
+backend.
```
mgeisler, line 10: Capitalization is always a difficult topic... here I looked at https://keats.github.io/tera/docs/ and see that the project refer to themselves as "Tera":

```suggestion
This `mdbook` backend makes it possible to use [Tera](https://github.com/Keats/tera)
templates and expand the capabilities of your books. It works on top of the default HTML
backend.
```

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::tera_renderer::custom_component::TeraRendererConfig;
+use crate::tera_renderer::renderer::Renderer;
+
+/// Re-renders HTML files outputed by the HTML backend with Tera templates.
+/// Please make sure the HTML backend is enabled.
+fn main() -> anyhow::Result<()> {
+    let mut stdin = io::stdin();
+    let ctx = RenderContext::from_json(&mut stdin).unwrap();
+    if ctx.config.get_preprocessor("html").is_none() {
```
mgeisler, line 15:
```suggestion
    if ctx.config.get_renderer("html").is_none() {
```

mgeisler: This tells me that we need a integration test for this :slightly_smiling_face:

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+### Using templates in `index.hbs`
+
+Since the HTML renderer will first render Handlebars templates, we need to tell it to
+ignore tera templates using `{{{{raw}}}}` blocks:
+
+```html
+{{{{raw}}}}
+{% set current_language = ctx | get(key="config") | get(key="book") | get(key="language", default="en") %}
```
mgeisler, line 57: This can be simplified to just

```suggestion
{% set current_language = ctx.config.book.language %}
```

If you want to handle an unset `language`, the you can do

```
{% set current_language = ctx.config.book.language | default(value="en") %}
```

However, [`mdbook` uses `"en"` by default](https://github.com/rust-lang/mdBook/blob/94e0a44e152d8d7c62620e83e0632160977b1dd5/src/config.rs#L427) (but wraps it in `Some`).

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+        return Err(anyhow!(
+            "Could not find the HTML backend. Please make sure the HTML backend is enabled."
+        ));
+    }
+    let config: TeraRendererConfig = ctx
+        .config
+        .get_deserialized_opt("output.tera-backend")
+        .expect("Failed to get tera-backend config")
+        .unwrap();
```
mgeisler, line 24: You can avoid the `unwrap` with an extra `?` (you need to import `anyhow::Context`):

```suggestion
    let config: TeraRendererConfig = ctx
        .config
        .get_deserialized_opt("output.tera-backend")
        .context("Could not load output.tera-backend config")?
        .ok_or(anyhow!("Missing output.tera-backend config"))?;
```

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let tera_template = config
+        .create_template(&ctx.root)
+        .expect("Failed to create components");
```
mgeisler, line 28:
```suggestion
    let tera_template = config.create_template(&ctx.root)?;
```

mdbook-tera-backend/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let mut renderer = Renderer::new(ctx, tera_template).expect("Failed to create renderer");
+
+    renderer.render_book().expect("Failed to render book");
+
+    Ok(())
```
mgeisler, line 34:
```suggestion
    let mut renderer = Renderer::new(ctx, tera_template)?;
    renderer.render_book()
```

mgeisler: Avoid using `expect` when you have "proper" error handling available via `anyhow`. If the default error message isn't good enough, you can tweak it with `.context(...)` (but here I'm guessing it's good enough so we can just do the simplest thing possible.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// by considering all the output HTML files as input to the Tera template.
+    /// It overwrites the preexisting files with their Tera-rendered version.
+    pub(crate) fn render_book(&mut self) -> Result<()> {
+        let dest_dir = self
+            .ctx
+            .destination
+            .parent()
+            .unwrap()
+            .join("html")
+            .to_owned();
```
mgeisler, line 60: No `to_owned` needed, `join` already return you an owned value:

```suggestion
        let dest_dir = self.ctx.destination.parent().unwrap().join("html");
```

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+            return Err(anyhow!(
+                "{dest_dir:?} is not a directory. Please make sure the HTML renderer is enabled."
+            ));
+        }
+        self.render_book_directory(&dest_dir)
+    }
+
+    /// Create the `get_context` function for the `Tera` template, a helper that allows retrieving values from `ctx`.
+    fn create_get_context_function(&self) -> impl tera::Function {
+        let ctx_rc = Arc::clone(&self.ctx);
```
mgeisler, line 71: It's not idiomatic to encode the type ("reference counted") into the variable name:

```suggestion
        let ctx = Arc::clone(&self.ctx);
```

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+        move |args: &HashMap<String, serde_json::value::Value>| -> tera::Result<tera::Value> {
+            let key = args
+                .get("key")
+                .ok_or_else(|| tera::Error::from(format!("No key argument provided")))?
```
mgeisler, line 75: Please use [`Error::msg`](https://docs.rs/tera/latest/tera/struct.Error.html#method.msg) instead:

```suggestion
                .ok_or_else(|| tera::Error::msg("No key argument provided"))?
```

Also twice below.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Reads the file at `path` and renders it.
+    fn process_file(&mut self, path: &Path) -> Result<()> {
+        if path.extension().unwrap_or_default() != "html" {
+            return Ok(());
+        }
+        let file_content = std::fs::read_to_string(path)?;
+        let output = self.render_file_content(&file_content, path)?;
+        let mut output_file = fs::File::create(path)?;
+        output_file.write_all(output.as_bytes())?;
+        Ok(())
```
mgeisler, line 112: Please try with [std](https://doc.rust-lang.org/stable/std/index.html)::[fs](https://doc.rust-lang.org/stable/std/fs/index.html)::[write](https://doc.rust-lang.org/stable/std/fs/fn.write.html#):

```suggestion
        fs::write(path, output)
```

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// `path`: The path to the file that will be added as extra context to the renderer.
+    fn create_context(&mut self, path: &Path) -> Result<tera::Context> {
+        let mut context = tera::Context::new();
+        context.insert("path", path);
+        context.insert("ctx", &serde_json::to_value(&*self.ctx)?);
+        context.insert("book_dir", &self.ctx.destination.parent().unwrap());
+        context.insert("attributes", &BTreeMap::<String, String>::new());
```
mgeisler, line 125: This is not used from what I can tell?

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Creates the rendering context to be passed to the templates.
+    ///
+    /// # Arguments
+    ///
+    /// `path`: The path to the file that will be added as extra context to the renderer.
+    fn create_context(&mut self, path: &Path) -> Result<tera::Context> {
+        let mut context = tera::Context::new();
+        context.insert("path", path);
+        context.insert("ctx", &serde_json::to_value(&*self.ctx)?);
+        context.insert("book_dir", &self.ctx.destination.parent().unwrap());
```
mgeisler, line 124: This can be computed from `ctx` so it looks unnecessary?

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+        if !dest_dir.is_dir() {
+            return Err(anyhow!(
+                "{dest_dir:?} is not a directory. Please make sure the HTML renderer is enabled."
+            ));
+        }
+        self.render_book_directory(&dest_dir)
+    }
+
+    /// Create the `get_context` function for the `Tera` template, a helper that allows retrieving values from `ctx`.
+    fn create_get_context_function(&self) -> impl tera::Function {
```
mgeisler, line 70: I'm not sure this is needed since you can just `ctx.foo.bar` in the template?

Removing this would most likely also allow you to remove the use of `Arc`.

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::sync::Arc;
+use tera::Tera;
+
+/// Renderer for the tera backend.
+///
+/// This will read all the files in the `RenderContext` and render them using the `Tera` template.
+///
+/// # Example
+///
+/// ```
```
mgeisler, line 17: Because this is a binary crate (there is no `src/lib.rs` file), the doctests are actually not executed. This is a problem since it gives you (and me! :smile:) a false sense of security.

Please move the doctests to unit or integration tests so that we can have confidence that the main parts of the code are tested. Right now, I see this

```
% cargo test
   Compiling mdbook-tera-backend v0.0.1 (.../mdbook-i18n-helpers/mdbook-tera-backend)
    Finished test [unoptimized + debuginfo] target(s) in 6.87s
     Running unittests src/main.rs (.../mdbook-i18n-helpers/target/debug/deps/mdbook_tera_backend-33d75d56994baca7)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

mdbook-tera-backend/src/tera_renderer/renderer.rs
sakex: That's annoying to compute from `ctx` in tera. I think it's a good addition

sakex:
@mgeisler tests added

mdbook-tera-backend/src/tera_renderer/renderer.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[test]
+    fn test_renderer() {
+        let mut ctx = RenderContext::from_json(RENDER_CONTEXT_STR.as_bytes()).unwrap();
+
+        let tmp_dir = TempDir::new("output").expect("Failed to create temp dir");
+        let html_path = tmp_dir.path().join("html");
+        let templates_path = tmp_dir.path().join("templates");
+
+        std::fs::create_dir(&html_path).expect("Failed to create html dir");
+        std::fs::create_dir(&templates_path).expect("Failed to create html dir");
```
mgeisler, line 150: Please drop the `.expect()` calls (because they're using the wrong strings). Just use `?` and let your test return a `Result`.

sakex: I disagree on that, the error message is really not useful if we return an anyhow result, even with .context().

mgeisler: It's a test -- if it fails, then the developer will have to go and debug it by hand. So we don't need to make this "user friendly" since the user is a developer and since they will have to debug it anyway.

mdbook-tera-backend/README.md
mgeisler: You resolved this, but I still see a lot of `ctx | get(key="config") | get(key="book")` above? Why is that?

mdbook-tera-backend/src/tera_renderer/renderer.rs
mgeisler: It can be computed with a small pipeline:

```jinja
{{ ctx.destination | split(pat="/") | slice(end=-1) | join(sep="/") }}
```

mgeisler (changes requested):

mgeisler:
> @mgeisler tests added

Thank you, this helps a lot in the confidence as we refactor this going forward.

mdbook-tera-backend/src/tera_renderer/renderer.rs
sakex: But what's the advantage of returning a result here, if it does the same thing with worse error messages?

mgeisler: > But what's the advantage of returning a result here, if it does the same thing with worse error messages?

My thinking is that it'll lead to less code in the tests. Test have to be _super clear_ since they're the thing that keep the of the code working :smile: So every line there is a liability — there is nothing to check or validate the line except you and me.

In this case, the code was

```rust
        std::fs::create_dir(&html_path).expect("Failed to create html dir");
        std::fs::create_dir(&templates_path).expect("Failed to create html dir");
```
which is wrong: the error messages are the same but they do different things. So that's a problem and the problem can be avoided (by construction) if you write less code with a `?`. Then you cannot make this mistake at all.

The test will , it will only break when someone is refactoring the code in question. So that person will already be in the middle of a development cycle and so having the test fail is hopefully be met with "Ah, yes of course! I'll have to adjust this now!"

mgeisler: The error messages are also off below:

```rust
        std::fs::write(templates_path.join("test_template.html"), TEMPLATE_FILE)
            .expect("Failed to create test.html");
```

and this just shows that it's annoying to keep things in sync :smile: So having less things to keep in sync should be an overall win for everbody.

mgeisler: One of our internal [Testing on the Toilet](https://testing.googleblog.com/2007/01/introducing-testing-on-toilet.html) episodes touched upon this: tests should avoid complexity at all cost. They should be obviously correct and don't try to factor common code out into helpers and constants.

So I would even suggest removing the small string constants above and just inline the content below.

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+### Using templates in `index.hbs`
+
+Since the HTML renderer will first render Handlebars templates, we need to tell
+it to ignore Tera templates using `{{{{raw}}}}` blocks:
+
+```html
+{{{{raw}}}} {% set current_language = ctx.config.book.language %}
+<p>CURRENT LANGUAGE: {{ current_language }}</p>
```
mgeisler, line 55:
```suggestion
{{{{raw}}}}
{% set current_language = ctx.config.book.language %}
<p>Current language: {{ current_language }}</p>
```

mdbook-tera-backend/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Creating templates
+
+Create your template files in the same directory as your book.
+
+```html
+<!-- ./theme/templates/language_list.html -->
+<ul>
+  {% for identifier, language_name in get_context(key="output.i18n.languages")
+  %}
```
mgeisler, line 42: Is `get_context` still there? I think it got removed and that this can be replaced with

```suggestion
  {% for identifier, language_name in get_context(key="ctx.config.output.i18n.languages") %}
```

I don't think you show the corresponding `book.toml` for this anywhere? This makes the example a bit complicated in my opinion.

mgeisler (approved):
Hey @sakex, I think this has been cooking more than needed now!

Thank you for fixing many small thing. From my testing, I can see that this is a huge step forward! It allows us to solve some very long-standing and important problems in Comprehensive Rust (as shown in https://github.com/google/mdbook-i18n-helpers/issues/35#issuecomment-1793750265).

mdbook-tera-backend/README.md
mgeisler: We can fix this before/after merging this PR — we'll have time to go over the documentation before we release this on crates.io.

Issue #57, opened 2023-08-15
Add a fuzz tests for all binaries


mgeisler:
We should add fuzz tests for each binary:

- `mdbook-xgettext`: test that we don't crash on random Markdown input
- `mdbook-gettext`: test that we don't crash on random PO files
- `mdbook-i18n-normalize`: same as for `mdbook-gettext` (this could have caught #56)

kdarkhan:
Hey, I would like to work on this issue.

kdarkhan:
Hi @mgeisler,

Sorry, been busy at my main work lately. I started looking into this issue. I will try to create a PR in the next few days.

I had some questions:

1. This task is about fuzzing binaries. One approach I can do is within my fuzz test execute the binary by using fuzz provided random input as a file and executing the binaries against that. That probably will be slow and instead I should fuzz main functions in the binaries. Is that correct?
2. I wanted to start with fuzzing with a sample fuzz target that catches the error fixed by https://github.com/google/mdbook-i18n-helpers/pull/56. Did that PR fix a wrong output or a panic? That is in my fuzz test for `mdbook-i18n-normalize` - should I be testing equality for some specific expected output or just that the function does not panic?

mgeisler:
Thanks for looking at this! I completely understand that main work can become busy :slightly_smiling_face: 

> 1. This task is about fuzzing binaries. One approach I can do is within my fuzz test execute the binary by using fuzz provided random input as a file and executing the binaries against that. That probably will be slow and instead I should fuzz main functions in the binaries. Is that correct?

Yeah, you should fuzz the main logic by calling functions from your fuzz tests — you should not execute the binaries. I would not focus on the binaries, just use them as a starting point to find the logic to fuzz. Find the functions called by `main` and try to restructure the code so that it can be fuzzed. This likely means moving a bit more functionality from `src/bin/...` to `src/lib.rs`. If we start seeing related functionality in `lib.rs`, then we'll move it into separate Rust modules.

> 2. I wanted to start with fuzzing with a sample fuzz target that catches the error fixed by [Handle normalization where old `msgid` disappears #56](https://github.com/google/mdbook-i18n-helpers/pull/56). Did that PR fix a wrong output or a panic? That is in my fuzz test for `mdbook-i18n-normalize`, should I be testing equality for some specific expected output or just that the function does not panic?

[...]

PR #81, opened 2023-09-15
Implement  fuzzer for normalize

kdarkhan:
This addresses https://github.com/google/mdbook-i18n-helpers/issues/57.

I am focusing on `mdbook-i18n-normalize` binary only in this PR. I will implement fuzzers for other binaries separately.
In order to fuzz the logic, I had to move it out of binary to library. The first commit
copies the binary file without making any changes to make the review easier.

I confirmed that when running the new fuzzer against the version of the code
before https://github.com/google/mdbook-i18n-helpers/pull/56 was merged,
it finds a failure pretty quickly.


mgeisler:
> In order to fuzz the logic, I had to move it out of binary to library. The first commit
> copies the binary file without making any changes to make the review easier.

Nice, thanks for thinking of your reviewer! 🙂

Please squash the first two commits into one so that you _move_ the code instead of copying and deleting it. The way I see it, the repository should look good at any commit and ir someone where to checkout the first commit, they would probably wonder about the duplicate code 

Also, please name the new module `nornalize`, I think the `i18n_` prefix is redundant since we're already inside a `mdbook_i18n_helpers` module.

kdarkhan:
Sounds good. I removed the prefix `i18n_` from the module name and the fuzz target name, squashed the commits as well.

src/normalize.rs
```diff
@@ -0,0 +1,548 @@
+//! Normalize the Markdown in a  a PO or POT file.
```
mgeisler, line 1: Tiny typo:

```suggestion
//! Normalize the Markdown in a PO or POT file.
```

mgeisler:
Awesome, this looks great!

mgeisler:
@kdarkhan, would #83 be something for you since you're already looking at fuzzers?

kdarkhan:
@mgeisler, sure, I will work #83 as well once I am done with the other 2 fuzz tests.

mgeisler:
Thanks!

Issue #18, opened 2023-02-16
Add scripts for using and updating translations


mgeisler:
The instructions for our translation pipeline in [`TRANSLATIONS.md`](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md) don't match the actual steps in [`publish.yml`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml).

The difference is mostly because of how the GitHub Actions allows us to set environment variables using a different syntax.

I would like to unify the two via small scripts. The scripts could be shell scripts (though that probably doesn't work well on Windows?) or they could be Rust "scripts" (more setup time).

I'm imagining something like

* `build-translation` which takes a `xx` locale and outputs a book in `book/xx`.
* `update-translation` which runs both `mdbook-xgettext` and `msgmerge` for you.

A `serve-translation` would probably also be nice to have.

Instead of several scripts, a single script with subcommands could also be nice. That could probably live nicely in the `i18n-helpers` project since it would be tightly coupled to the other binaries there.

mgeisler:
(This issue was transferred from https://github.com/google/comprehensive-rust/.)

mgeisler:
The GitHub action in #48 could call into this script and so we avoid putting too much logic into the action (GitHub actions are nice in the context of GitHub, but they're hard to run by hand on your local machine).

jonodeno:
@mgeisler I would be interested in taking this up. I would do it as a Rust script and I can start with `build-translation`. Just to be sure that I'm understanding this correctly we would have:
- build-translation
    - runs `mdbook-build` 
    - runs `msginit -i book/xgettext/messages.pot -l xx -o po/xx.po`
    - make the change to `.github/labeler.yml` like so:
```
+"translation/xx":
+  - changed-files:
+      - any-glob-to-any-file: po/xx.po
```

mgeisler:
Hi @jonodeno, thanks for looking at this and sorry for the late reply!

At the moment, we use [`build.sh`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/build.sh) in Comprehensive Rust to render a single language. So this script shows the kind of functionality I'm looking for as a baseline.

The full script should then do this for every language.

See also #84 by @sakex, which was an early attempt at doing this.

PR #84, opened 2023-09-22
Build translations from mdbook

sakex:


sakex:
We should bump the version as alpha in `cargo.toml` and publish the new version

mgeisler:
This basically moves the `for` loop from [`publish.yml`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml) to the Rust code?

Later I guess this for-loop-renderer (🙂) will know to pass in i18n-releated information to the book — or maybe not since the other renderer will already be able to grab this from the `book.toml` file?

src/main.rs
mgeisler, line 1: I'm not super keen on indirectly reserving `mdbook-i18n-helpers` as the name for the mdbook backend via a file named `main.rs`. I think I would be happier with `src/bin/mdbook-i18n.rs`, which would allow us to use `i18n` as the renderer name.

It would then be a conflict to install https://github.com/funkill/mdbook-i18n at the same time. I hope we can get away with that — if not, then `mdbook-i18-helper` or a similar singular name would be better.

sakex:
> This basically moves the `for` loop from [`publish.yml`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml) to the Rust code?

Yes that's the only thing it does. The advantage is that we get a better developer experience for translators. They see all translations at the same time without having to rebuild multiple times with a different command. Live reload is also supported. It's a small win but I think it's worth it.

src/main.rs
sakex: Sounds good, I'm moving the file

Cargo.toml
```diff
@@ -17,7 +17,9 @@ pulldown-cmark = { version = "0.9.2", default-features = false }
 pulldown-cmark-to-cmark = "10.0.4"
 regex = "1.9.4"
 semver = "1.0.16"
+serde = "1.0.130"
 serde_json = "1.0.91"
+toml = "0.5.8"
```
mgeisler, line 22: Could you add a comment saying that we need to stick to the version used by `mdbook`? Because of https://github.com/rust-lang/mdBook/pull/2151. Ideally, I would like to avoid having to know this by not using `toml` directly, like we did in https://github.com/google/mdbook-i18n-helpers/pull/77#discussion_r1325560711.

However, I'm realizing that this seems like a loosing battle, so we can probably depend on it and then rely on the compiler telling us when things cannot be mixed.

src/main.rs
```diff
@@ -0,0 +1,83 @@
+use mdbook::renderer::RenderContext;
+use serde::Deserialize;
+use std::collections::BTreeMap;
+use std::io;
+
+#[derive(Deserialize)]
+struct I18nConfiguration {
```
mgeisler, line 7: We need documentation comments for the new structs. Please explain what the indented use of them are.

src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+fn main() {
+    let mut stdin = io::stdin();
+
+    // Get the configs
+    let ctx = RenderContext::from_json(&mut stdin).unwrap();
+    let i18n_config: I18nConfiguration = ctx
+        .config
+        .get_deserialized_opt("output.i18n-helpers")
```
mgeisler, line 24: There is a [get_renderer](https://docs.rs/mdbook/latest/mdbook/config/struct.Config.html#method.get_renderer) method available, but I guess it doesn't improve much?

src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+        .unwrap()
+        .unwrap();
```
mgeisler, line 26: Not really your fault... but how does the error look if someone enables the renderer but doesn't fill out any of the values (I guess `languages` and `default_language` are required values?)

sakex: Thanks for pointing out, I added `serde(default)` on languages and now, it will just have empty values

src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+    // Overwrite with current values from stdin.
+    mdbook.book = ctx.book.clone();
+    mdbook.config = ctx.config.clone();
+    mdbook.root = ctx.root.clone();
+
+    let book_config = mdbook
+        .config
+        .get_mut("output.i18n-helpers")
+        .expect("No output.i18n-helpers config in book.toml");
+    // Set translate_all_languages to false for nested builds to prevent infinite recursion.
```
mgeisler: I wonder what the use case if for having `translate_all_languages`? Could we not decide that it is a configuration mistake to enable the `i18n` backend if you don't want to translate all languages? So just write the code as if it was always true.

src/main.rs
sakex: It returns a `HashMap`, it's nicer to have a typed `struct`

src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+    book_config
+        .as_table_mut()
+        .expect("output.i18n-helpers config in book.toml is not a table")
+        .insert(
+            String::from("translate_all_languages"),
+            toml::Value::Boolean(false),
+        );
+
+    let output_directory = ctx.destination.parent().unwrap();
```
mgeisler: Here you're crawling out of [`ctx.destination`](https://docs.rs/mdbook/latest/mdbook/renderer/struct.RenderContext.html#structfield.destination), which says

> Where the renderer _must_ put any build artefacts generated.

Instead of relying on `output_directory.join("html")` below, I would suggest the following:

1. Run the `html` renderer somehow and make it output to `ctx.destination`.
2. Run it for each `lang` and make it output to `ctx.destination.join(lang)`.

The advantage: no implicit dependency on people enabling the `html` renderer (I believe it's allowed to disable it).

src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+        .get_deserialized_opt("output.i18n-helpers")
+        .unwrap()
+        .unwrap();
+
+    if !i18n_config.translate_all_languages {
+        return;
+    }
+
+    let mut mdbook = mdbook::MDBook::load(&ctx.root).expect("Failed to load book");
+    // Overwrite with current values from stdin.
```
mgeisler: When I read this, I wonder how the values can be different? Perhaps explain why this is necessary/useful in the comment?

src/main.rs
mgeisler: Thanks!

src/main.rs
sakex: For performance, if you have many translations, and a potato computer, it could be slow

src/main.rs
sakex: I rewrote that part a bit. That's why we have `move_translations_to_html_directory` also

src/main.rs
sakex: I rewrote that part a bit. That's why we have move_translations_to_html_directory also

Cargo.toml
sakex: Ideally, they would re-export their version

sakex: `.into()` did the trick x)

mgeisler:
A question: I imagine that we would want a `book.toml` that looks like this:

```
[output.i18n]
```

This enabled the `i18n` backend and _disables the `html` backend_ (as per the [documentation](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#output-tables)).

I imagine most projects will enable just this one backend and thus get the generated output put into `book/` by default.

In that setup, what does the `translate_all_languages` setting do? When `translate_all_languages` is off, do I get the same output into `book/` as if I had `[output.html]` enabled instead? If `translate_all_languages` is on, then I get `book/` as before _plus_ `book/xx/` for each locale?

I would like to see this described somewhere. Probably in the docstring for the new binary at first — later we can move this to a README for the new crate so that we can publish it on crates.io.

src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,110 @@
...
+    languages: BTreeMap<String, String>,
+    /// Default language code. It will not be translated.
+    default_language: Option<String>,
+
+    /// Whether to translate all languages or just the selected language, defaults to false.
+    #[serde(default)]
+    translate_all_languages: bool,
+    /// Whether to move the translations to the html directory, defaults to false.
+    ///
+    /// By default, translations's HTML output will live in `book/i18n/<language>/html`. If this is set to true, they will be moved to `book/html/<language>`.
```
mgeisler, line 32: I think the `book/i18n/` directory here is wrong: mdbook will only create a per-renderer subdirectory if there are multiple renderers enabled. We should no assume that `html` or any other renderer is enabled.

mgeisler: My suggestion would be to create a tiny blank book and experiement with the new renderer there. Don't use Comprehensive Rust as a starting point — since that book has a more complex setup than most (I think).

sakex:
> A question: I imagine that we would want a `book.toml` that looks like this:
> 
> ```
> [output.i18n]
> ```
> 
> This enabled the `i18n` backend and _disables the `html` backend_ (as per the [documentation](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#output-tables)).
> 
> I imagine most projects will enable just this one backend and thus get the generated output put into `book/` by default.
> 
> In that setup, what does the `translate_all_languages` setting do? When `translate_all_languages` is off, do I get the same output into `book/` as if I had `[output.html]` enabled instead? If `translate_all_languages` is on, then I get `book/` as before _plus_ `book/xx/` for each locale?
> 
> I would like to see this described somewhere. Probably in the docstring for the new binary at first — later we can move this to a README for the new crate so that we can publish it on crates.io.

We do need to have the HTML renderer to generate HTML output. It is also documented that we can simply set `[output.html]` in `book.toml`. We should mention that in the doc. They mention it in [mdbook-pdf](https://crates.io/crates/mdbook-pdf) for instance. 

Since the preprocessor only edits the content of the book object but doesn't create any files itself, it will work with any other renderer as well, which makes me think that we could change the parameter from `move_translations_to_html_directory: bool` to `move_translations_to_renderer_outputs: Vec<String>` which would take arguments like `html` and move the translated output from `$lang/$renderer/` to `$renderer/$lang`

mgeisler:
> We do need to have the HTML renderer to generate HTML output. It is also documented that we can simply set `[output.html]` in `book.toml`. We should mention that in the doc. They mention it in [mdbook-pdf](https://crates.io/crates/mdbook-pdf) for instance.

Ah, I guess that makes sense: people should configure the HTML output via the normal `html` renderer settings! I was thinking we would remove this, but it's not a good idea.

> Since the preprocessor only edits the content of the book object but doesn't create any files itself, it will work with any other renderer as well, which makes me think that we could change the parameter from `move_translations_to_html_directory: bool` to `move_translations_to_renderer_outputs: Vec<String>` which would take arguments like `html` and move the translated output from `$lang/$renderer/` to `$renderer/$lang`

I like the idea of making this generic over the kind of renderer! As a litmus test, you could ensure that it's possible to generate i18n output for the built-in `markdown` renderer.

Now that discuss this: I think the correct strategy would be to copy the output _from_ the other renderer and into the `i18n` renderer's output directory. This avoids writing anywhere except for `ctx.destination` and it is thus consistent with how mdbook renderers are supposed to work.

From skimming [Alternative Backends](https://rust-lang.github.io/mdBook/for_developers/backends.html), I don't see any ordering between the renderers. Do we rely on the existing HTML output being present? If we do, we should probably not do this and generate it ourselves :slightly_smiling_face:

sakex:
> From skimming [Alternative Backends](https://rust-lang.github.io/mdBook/for_developers/backends.html), I don't see any ordering between the renderers. Do we rely on the existing HTML output being present? If we do, we should probably not do this and generate it ourselves 🙂

In this renderer, we don't care about what runs first. For the new template renderer, we will need html to be present. The rendering order is based on the order in which they are defined in `book.toml`. I remember seeing it in the doc, but I can't find it anymore, reading the code, it is how it is handled.

mgeisler:
Just a note about _Merge branch 'main' into translate-from-mdbook_. Please rebase your commits on top of `main` instead of merging `main` into the branch. By rebasing, you create a clean and useful history — by merging, you end up with a mix of your changes and other changes and it's hard to read the history afterwards.

For this reason, I will want to squash merge the PR if it has merge commits.

mgeisler:
> > From skimming [Alternative Backends](https://rust-lang.github.io/mdBook/for_developers/backends.html), I don't see any ordering between the renderers. Do we rely on the existing HTML output being present? If we do, we should probably not do this and generate it ourselves 🙂
> 
> In this renderer, we don't care about what runs first.

We don't care because we run the `html` renderer ourselves?

> For the new template renderer, we will need html to be present.

Could you instead run the `html` renderer yourself? Output the files to your `ctx.destination` and the process each of them through the template engine.

> The rendering order is based on the order in which they are defined in `book.toml`. I remember seeing it in the doc, but I can't find it anymore, reading the code, it is how it is handled.

Okay, thanks for checking! If it isn't documented, then we should not rely on it. If the order is important, could you then open an PR to the upstream repository to define the order in the documentation?

sakex:
> Just a note about _Merge branch 'main' into translate-from-mdbook_. Please rebase your commits on top of `main` instead of merging `main` into the branch. By rebasing, you create a clean and useful history — by merging, you end up with a mix of your changes and other changes and it's hard to read the history afterwards.
> 
> For this reason, I will want to squash merge the PR if it has merge commits.

Yes, I did that because it is more convenient. I only want one commit for this PR anyway, so squashing is expected

sakex:
Hey @mgeisler . Is it Ok to merge now (with squash) so that I can use this as a base for the other renderer?

mgeisler:
> Hey @mgeisler . Is it Ok to merge now (with squash) so that I can use this as a base for the other renderer?

I think we need some documentation before we merge this. It's not clear to me what the expectations are for the user.

As an example, I believe you said that the `html` backend has to be enabled?

> After the renderer is done, I'll make a test book to test all the features

Please add a test here already, otherwise it's hard to know what to expect. Also, update the first PR comment above to explain what this does (hint: reuse that text for your module docstring and for the README and USAGE).

mgeisler:
> > > From skimming [Alternative Backends](https://rust-lang.github.io/mdBook/for_developers/backends.html), I don't see any ordering between the renderers. Do we rely on the existing HTML output being present? If we do, we should probably not do this and generate it ourselves 🙂
> > 
> > 
> > In this renderer, we don't care about what runs first.
> 
> We don't care because we run the `html` renderer ourselves?

We still need to answer this essential question: how can we rely on `ctx.destination + "/../html"` existing when this backend is executed? I believe we still have that implicit expectation?

sakex:
> > > > From skimming [Alternative Backends](https://rust-lang.github.io/mdBook/for_developers/backends.html), I don't see any ordering between the renderers. Do we rely on the existing HTML output being present? If we do, we should probably not do this and generate it ourselves 🙂
> > > 
> > > 
> > > In this renderer, we don't care about what runs first.
> > 
> > We don't care because we run the `html` renderer ourselves?
> 
> We still need to answer this essential question: how can we rely on `ctx.destination + "/../html"` existing when this backend is executed? I believe we still have that implicit expectation?

No, it's not required anymore

i18n-helpers/src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        let translation_path = output_directory.join(language);
+
+        // Book doesn't implement clone, so we just mutate in place.
+        mdbook.config.book.language = Some(language.clone());
+        mdbook.config.book.multilingual = true;
+        mdbook.config.build.build_dir = translation_path;
+        mdbook
+            .build()
+            .unwrap_or_else(|_| panic!("Failed to build translation for language: {}", language));
```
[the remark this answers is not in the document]
mgeisler: Can you let `main` return `anyhow::Result<()>` and then use `with_context` here instead? That should let you remove most of the scary calls to `unsafe` above and below.

i18n-helpers/src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,107 @@
...
+        for renderer in &i18n_config.move_translations_directories {
+            std::fs::create_dir_all(output_directory.parent().unwrap().join(renderer))
+                .unwrap_or_else(|_| panic!("Failed to create html directory in output directory"));
```
mgeisler, line 93: The error message if off and this seems to be the place where you violate the contact of a mdbook renderer?

This also seems to be where you depend on the `html` renderer running before this one?

My guess is that there can be bugs hiding here: such as accidentally copying the _previous_ output of `mdbook build`?

mgeisler (changes requested):
Hey @sakex, I tried marking the things we need to figure out or explain before we merge this.

i18n-helpers/src/bin/mdbook-i18n.rs
sakex: I forgot to replace the message when I changed the code here

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
sakex: Good idea, let's do that. Panic is not unsafe though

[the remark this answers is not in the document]
sakex: It's clippy that transformed the `expect(format!` to `unwrap_or_else(panic`

i18n-helpers/src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    if !i18n_config.translate_all_languages {
+        return;
+    }
+
+    let mut mdbook = mdbook::MDBook::load(&ctx.root).expect("Failed to load book");
+    // Overwrite with current values from stdin. This is necessary because mdbook will add data to the config.
+    mdbook.book = ctx.book.clone();
+    mdbook.config = ctx.config.clone();
+    mdbook.root = ctx.root.clone();
```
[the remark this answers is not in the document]
sakex: Good point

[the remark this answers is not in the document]
sakex: I'm not home today, so I won't be able to fix those until tomorrow

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: > Panic is not unsafe though

Sorry, I meant to say `unwrap` above!

> It's clippy that transformed the `expect(format!` to `unwrap_or_else(panic`

Yeah, that's really annoying from a code clarity point of view. I hope it can be fixed with more use of `?` to just return early.

i18n-helpers/src/bin/mdbook-i18n.rs
mgeisler: > This also seems to be where you depend on the `html` renderer running before this one?
> 
> My guess is that there can be bugs hiding here: such as accidentally copying the _previous_ output of `mdbook build`?

The questions here are the important part: we have to avoid getting ourselves into a TeX-like situation where you need to run `mdbook build` twice to get the correct final output :smile:

i18n-helpers/src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Whether to translate all languages or just the selected language, defaults to false.
+    #[serde(default)]
+    translate_all_languages: bool,
+    /// Whether to move the translations to the html directory, defaults to false.
+    ///
+    /// By default, translations' output will live in `book/i18n/<language>/<renderer>`.
+    /// For all renderers in this list, we will move individual translations to `book/<renderer>/<language>`.
+    #[serde(default)]
+    move_translations_directories: Vec<String>,
```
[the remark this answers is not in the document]
sakex: It's very convenient with `mdbook serve` to view the output like it will be on the final site. It doesn't need to be enabled otherwise and by default does nothing

i18n-helpers/src/bin/mdbook-i18n.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// "es" = "Spanish (Español)"
+    /// "ko" = "Korean (한국어)"
+    /// "pt-BR" = "Brazilian Portuguese (Português do Brasil)"
+    /// ```
+    #[serde(default)]
+    languages: BTreeMap<String, String>,
+    /// Default language code. It will not be translated.
+    default_language: Option<String>,
+
+    /// Whether to translate all languages or just the selected language, defaults to false.
```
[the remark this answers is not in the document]
sakex: The language set in the env variable `MDBOOK_BOOK__LANGUAGE`

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: >> I think you're saying here that you'll [_break the contract_ defined by mdbook](https://rust-lang.github.io/mdBook/for_developers/backends.html#output-and-signalling-failure).
>> 
>> I don't want the new renderer to touch files in the output directories of other renderers.
>
> It's very convenient with `mdbook serve` to view the output like it will be on the final site. It doesn't need to be enabled otherwise and by default does nothing

I don't think you addressed my concern: if the code is writing outside of the output directly, then the code is doing something wrong.

I don't doubt that it is convenient: we should just write everything inside of the designated output directory.

Maybe you're not aware, but I often use `mdbook build -d /tmp/foo` and similar to generate output in non-standard directories. So this renderer must respect the output setting.

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: Ah, okay... Setting `MDBOOK_BOOK__LANGUAGE` is just another way of setting the `language` field in `book.toml`. So it's not strictly "selected", it's just what the config happens to be when the `mdbook-gettext` preprocessor is invoked.

The current system has the elegant approach that you can set `language` to what you want and it will then translate the text on the fly to this language — because it's a preprocessor which is independent of the renderer!

The new renderer is a bit different: it takes a list of languages from `book.toml` and renders the book for each of them. If the user only want to render a single language, the simplest way would be for the user to remove the other languages from the list in `book.toml`!

The users could do this on the command line with a suitable `MDBOOK_OUTPUT__I18n__LANGUAGES` setting (I think). This would be better since it uses less infrastructure: all you need to know is that the `output.i18n.languages` key controls the list of languages to generate for. If it has 1 item, then you get a single language, if it has $n$ items, then you get $n$ languages.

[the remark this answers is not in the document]
mgeisler: The conversion is documented here: https://rust-lang.github.io/mdBook/format/configuration/environment-variables.html.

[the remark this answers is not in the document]
sakex: The reason for this variable is that if you do  `MDBOOK_BOOK__LANGUAGE=fr mdbook build` with `translate_all_languages` to false, it will just translate `fr` otherwise, it will translate all languages. We can skip that, it's just for convenience.

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
sakex: By default, we don't break the contract but if we enable this, we do. That's also what we're already doing in comprehensive rust, so I don't really see a problem as it is opt-in only.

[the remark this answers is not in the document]
mgeisler: > That's also what we're already doing in comprehensive rust, so I don't really see a problem as it is opt-in only.

I hope our `mdbook` plugins are not breaking this?

Are you talking about [the code in `publish.yml`](https://github.com/google/comprehensive-rust/blob/77ab1ab6c41b03a62a09f22602da790b2af14717/.github/workflows/publish.yml#L60-L68), then that is quite different: the code there relies on the _contract of `mdbook`_ which says that it will put things into directory so-and-so. After calling `mdbook build`, we are of course free to move things around and mess with the `book/` directory —  we own it then!

This is the main difference between implementing this functionality as a wrapper script (as suggested in #18!) and writing it as a `mdbook` plugin. As a plugin, you are subjected to the rules of the surrounding system.

It might be a subtle difference, but I believe it's an important one. By following the principle carefully on every level of the stack, we will be able to build bigger "stacks" of software which behaves in a predictable manner when seen as a whole.

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: > The reason for this variable is that if you do `MDBOOK_BOOK__LANGUAGE=fr mdbook build` with `translate_all_languages` to false, it will just translate `fr` otherwise, it will translate all languages. We can skip that, it's just for convenience.

Can we unify this behavior? We stop setting `MDBOOK_BOOK__LANGUAGE` and instead drive the behavior of the backend using _solely_ a new `output.i18n.languages` field. It should behave like this:

- The `book.toml` file checked in has all languages for the book.
- To build fewer languages, people can remove the unwanted languages from `book.toml`. Alternativley, they can use the environment variable support and run

  ```sh
  MDBOOK_OUTPUT__I18N__LANGUAGES='["fr"]' mdbook build
  ```

  to restrict the languages to just French.

Why have a list in `book.toml` instead of just using `po/*.po`? Because there could be extra meta data compared to what we find in the PO files. I'm thinking of things like `english_name = "French", name = "Français"` and maybe even `available_in_language_picker = true`.

Does that sounds like a good idea? So instead of a "all languages or a single language" based on `MDBOOK_BOOK__LANGUAGE` (which doesn't make a ton of sense for a multi-language book), we have a more flexible system that people can override on the command line.

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: > The script does this

Sorry, I don't think I know what script you're referring to here?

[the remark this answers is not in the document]
sakex: This renderer I mean

i18n-helpers/src/bin/mdbook-i18n.rs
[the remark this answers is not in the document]
mgeisler: > What you're suggesting would mean intercepting everything right at the beginning,

I'm not actually suggesting that — I was only suggesting changing the logic for which subset of the languages to build.

PR #86, opened 2023-09-22
Refactor repository to use workspaces

sakex:


mgeisler (approved):
Thanks, looks good!

mgeisler:
I think you just need to move the fuzz directory as well?

mgeisler (approved):
Thanks!

mgeisler:
We will need to either move the README as well, or create a new smaller README. Otherwise the mdbook-i18n-helpers crate will look odd when we publish it next.

Issue #16, opened 2023-04-04
Add support for only publishing a language if it more than NN% translated


mgeisler:
When the source material keeps changing, the translations will naturally lag behind. In that case, it could be nice to only publish a new version if it mostly up-to-date, meaning it is more than NN% translated.

This kind of functionality can be built today by looking at the output of `msgfmt` since it shows the number of translated and untranslated messages. We should try to package the functionality in a reusable fashion.

mgeisler:
This is implemented in #87, https://github.com/google/comprehensive-rust/pull/1243 and https://github.com/google/comprehensive-rust/pull/1252!

mgeisler:
I'm implementing a slightly different scheme: instead of a percentage, I'm implementing support for freezing the English source for a translation. This means that translators can work on a PO file for weeks and fully complete a translations — changes to the English source text won't invalidate the translation any longer.

To update a translation, run `mdbook-xgettext` and `msgmerge` like normal: this will add new strings and also update the POT-Creation-Date field in the PO file. When publishing the translation next, the English source files are taken from the time of POT-Creation-Date. This will give translations much more control over when strings are invalidated.

/cc @henrif75.

mgeisler:
This was fixed via #87 and https://github.com/google/comprehensive-rust/pull/1243.

PR #87, opened 2023-09-24
Set POT-Creation-Date in newly generated POT files

mgeisler:
The data is added to PO files automatically by `msgmerge`. This will in turn be used when publishing a translation: it allows us to know which sources goes into a given PO file.

Part of #16.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,6 +418,7 @@ name = "mdbook-i18n-helpers"
 version = "0.2.3"
 dependencies = [
  "anyhow",
+ "chrono",
  "mdbook",
  "polib",
  "pretty_assertions",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ description = "Plugins for a mdbook translation workflow based on Gettext."
 
 [dependencies]
 anyhow = "1.0.68"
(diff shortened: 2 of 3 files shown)


mgeisler:
@sakex, I'm thinking we should surface this field in the template engine. That way we can write something like

> Published on `{{ now() }}`, translation current as of `{{ pot-creation-date }}`.

in the footer of each page.

sakex:
Sure, this can be done by adding the variable to the book object at the gettext preprocessor step, then we'll retrieve it from a custom template

mgeisler:
> Sure, this can be done by adding the variable to the book object at the gettext preprocessor step

Oh, interesting! Can a preprocessor modify the metadata passed onto later preprocessors and renderers? I was thinking it should become a context variable for the template engine.

sakex:
Yes actually, we could just pass time.now from the renderer On Sep 25, 2023 08:56, Martin Geisler ***@***.***> wrote:

Sure, this can be done by adding the variable to the book object at the gettext preprocessor step

Oh, interesting! Can a preprocessor modify the metadata passed onto later preprocessors and renderers? I was thinking it should become a context variable for the template engine.

mgeisler:
See https://github.com/google/comprehensive-rust/pull/1243 for the other half of this: updating the the data in POT-Creation-Date when publishing a book.

Issue #68, opened 2023-08-29
Create binaries to support stand-alone usage


mgeisler:
The `mdbook-xgettext` plugin can currently only be used from `mdbook` since it reads a special JSON format on standard input. However, it would be quite easy to use the same code to extract text from any Markdown file on disk. We should create a small binary which does this.

Similarly for `mdbook-gettext`: we should create a small binary which will translate a Markdown file into another language using a PO file.

Having these binaries would make it slightly easier to text our own code:

```shell
echo '> foo **bar** *baz*' | mdbook-i18n-xgettext
```

should be enough to produce a PO file on standard output (it would show that `*bar*` is normalized to `_bar_`, for example). I could have used this kind of ad-hoc testing when I was battling corner-cases in #33.

friendlymatthew:
Howdy @mgeisler -- I would like to work on this :) 

Just to clarify - we want to read an MD file from std input, normalize the markdown, and convert to a PO file.

- As I am new to the codebase, should I use `pulldown-cmark` to parse the MD or is there a preferred method? I looked for previous PR's but had a hard time looking for role models.

- I'm currently reading `mdbook-i18n-normalize.rs` as this one parses a MD file as well (but from a PO file). Once I create a catalog, I would `normalize(catalog)` and convert to a po file.


Please let me know if I am misunderstanding or need clarification!

mgeisler:
While building the translation machinery here, I realized that we don't need to tie it to mdbook. We are able to do the `.md -> .pot` text extraction on any Markdown file. And we are able to do the `.md + .po -> .md` translation on any pair of Markdown and PO file.

> Just to clarify - we want to read an MD file from std input, normalize the markdown, and convert to a PO file.

My thinking is even simpler: the extraction binary should do what `mdbook-xgettext` does, but operate on a file chosen by the user (or `stdin`).

As you can see, `mdbook-xgettext` calls `extract_messages`, which internally calls `pulldown-cmark` as you suggest. However, `extract_messages` is the central function that takes a Markdown files and find the translatable text (it knows to strip unimportant Markdown syntax, so `## Foo` becomes just `Foo` in the PO file). These lines are the ones I think you can copy to the new binary:

```rust
            for (lineno, msgid) in extract_messages(&chapter.content) {
                let source = format!("{}:{}", path.display(), lineno);
                add_message(&mut catalog, &msgid, &source);
            }
```

[...]

PR #92, opened 2023-09-27
markdown-xgettext - binary that converts markdown text to .po file

friendlymatthew:
Related to https://github.com/google/mdbook-i18n-helpers/issues/68

You can execute this binary via standard input or by specifying specific markdown file. 
> Usage: {prog_name} <input.md> <output.po> [-lang=<language> OR -lang <language>]

* Note: If -lang is not specified, we default to English.

Example command:

```
echo '> # Exercises

To practice your Async Rust skills, we have again two exercises for you:

* Dining philosophers: we already saw this problem in the morning. This time
  you are going to implement it with Async Rust.

* A Broadcast Chat Application: this is a larger project that allows you
  experiment with more advanced Async Rust features.

<details>

After looking at the exercises, you can look at the [solutions] provided.

[solutions]: solutions-afternoon.md

</details>
' | markdown-xgettext - output.po -lang=fr
```

Problems to address:
- export path: currently, we don't specify a specific path when we generate the .po file. @mgeisler mentioned this would be helpful for debugging. May I get more context please? 
- title path: how would you like me to specify? We can also do -title command line parsing.


i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+        \n\
+        The first paragraph about Foo.\n\
+        Still the first paragraph.*baz*\n";
+
+        let html_content = "<h1>Hello Foo! </h1>";
+
+        let mut md_file = File::create(&md_path)?;
+        let mut html_file = File::create(&html_path)?;
+        write!(md_file, "{}", markdown_content)?;
+        write!(html_file, "{}", html_content)?;
```
mgeisler, line 201: There is a [`std::fs::write`](https://doc.rust-lang.org/stable/std/fs/fn.write.html) shortcut for this.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        assert_eq!(validate_file(&md_path), true);
```
mgeisler, line 203: `cargo clippy` will most likely tell you to replace `assert_eq!(foo, true)` with `assert!(foo)` (and likewise for the `assert_eq!` below).

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+use std::{env, fs};
+
+fn split_argument(arg: &String) -> Option<String> {
+    let splitted: Vec<_> = arg.split('=').collect();
+    if splitted.len() == 2 {
+        return Some(splitted[1].to_string());
+    }
+
+    None
+}
```
mgeisler, line 17: I think it's time to use a real command line parsing library. Please take a look at how @djmitche did it in [mdbook-course](https://github.com/google/comprehensive-rust/blob/main/mdbook-course/src/main.rs).

.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 *~
+.idea/
```
mgeisler, line 3: I suggest adding this directory to a [global Git ignore file](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file) instead. Arguably, the `*~` line should also not be here since it's not related to this repository.

i18n-helpers/src/bin/markdown-gettext.rs
djmitche: (fwiw, I adapted that from the example in the mdbook docs)

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+
+fn build_catalog(lang_option: Option<String>) -> anyhow::Result<Catalog> {
+    match lang_option {
+        Some(lang_path) => {
+            let pot_path = PathBuf::from(&lang_path);
```
mgeisler, line 22: I believe `PathBuf` is only for when you need a _mutable buffer_. If you just need a read-only path, you should use `Path`.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+
+            if pot_path.exists() {
```
mgeisler, line 24: It is normally easier to just call the I/O function and let that one raise an error if something doesn't exist, cannot be read, is the wrong type, etc... 

Notice that the file could disappear under your feet in the split second between you calling `pot_path.exists()` and calling `po_file::parse(&pot_path)` below :slightly_smiling_face:  This implies that you need to somehow deal with the file not existing from the `parse` call — which you do nicely by returning an error.

Removeing the `if` statement will also save you a level of nesting and simplify the code a bit.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+                let catalog = po_file::parse(&pot_path)
+                    .map_err(|err| anyhow!("{err}"))
+                    .with_context(|| format!("Could not parse {:?} as PO file", pot_path))?;
+
+                return Ok(catalog);
+            }
+
+            bail!("--po must be specified")
+        }
+        None => bail!("--po must be specified"),
```
mgeisler, line 34: You should be able to remove the `match` with

```rust
let lang_option = lang_option.ok_or(anyhow!("--po must be specified"))?;
```

This will return early if the option is empty. See [`ok_or`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or) and [`anyhow!`](https://docs.rs/anyhow/latest/anyhow/macro.anyhow.html).

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+fn split_argument(arg: &String) -> Option<String> {
+    let splitted: Vec<_> = arg.split('=').collect();
+    if splitted.len() == 2 {
+        return Some(splitted[1].to_string());
+    }
+
+    None
+}
+
+fn build_catalog(lang_option: Option<String>) -> anyhow::Result<Catalog> {
```
mgeisler, line 19: I bet you only need read-only access to the string here, so make the signature `Option<&str>` to reflect this.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+) -> anyhow::Result<Vec<PathBuf>> {
+    let mut valid_files = Vec::new();
+
+    match (dir_option, file_option) {
+        (Some(_), Some(_)) => return bail!("Only one of --dir or --f should be specified"),
+        (None, None) => return bail!("Either --dir or --f must be specified"),
+        (Some(dir_path), None) => {
+            let full_dir = PathBuf::from(&dir_path);
+            if !full_dir.exists() || !full_dir.is_dir() {
+                return bail!("Directory does not exist: {}", full_dir.display());
```
mgeisler, line 84: [`bail!`](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) already has a `return` inside it:

```suggestion
                bail!("Directory does not exist: {}", full_dir.display());
```

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+            }
+
+            for entry in fs::read_dir(full_dir)? {
+                let entry = entry?;
+                let path = entry.path();
+
+                if validate_file(&path) {
+                    valid_files.push(path);
+                }
```
mgeisler, line 93: This looks like a [`filter`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.filter) operation.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,248 @@
...
+    match output_path_option {
+        Some(output_path) => {
+            let path = PathBuf::from(&output_path);
+            if path.exists() {
+                if !path.is_dir() {
+                    bail!(
+                        "Specified output path exists but is not a directory: {}",
+                        path.display()
+                    );
+                }
```
mgeisler, line 119: Like above, you should skip the error handling here and let `fs::create_dir_all` below raise an error if there is a problem.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        let output_file_name = file_path.file_name().ok_or(anyhow!("Invalid file name"))?;
+        let output_file_path = output_path.join(output_file_name);
+
+        let mut output_file = File::create(&output_file_path)
+            .with_context(|| format!("Failed to create file: {}", output_file_path.display()))?;
+
+        output_file
+            .write_all(translated_content.as_bytes())
+            .with_context(|| format!("Failed to write to file: {}", output_file_path.display()))?;
```
mgeisler, line 44: This can be `std::fs::write`, I think?

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,250 @@
...
+mod tests {
+    use super::*;
+    use tempfile::TempDir;
+
+    static MARKDOWN_CONTENT: &str = "# How to Foo from a specified file path\n\
+        \n\
+        The first paragraph about Foo.\n\
+        Still the first paragraph.*baz*\n";
+
+    static INVALID_CONTENT: &str = "<p> This is not markdown content </p>";
```
mgeisler, line 125: Actually... Markdown can contain HTML tags, so this string is also valid Markdown :slightly_smiling_face:

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+    let matches = Command::new("markdown-xgettext")
+        .about("binary that extracts translatable text from markdown file for mdbook-i18n-helpers")
+        .arg(Arg::new("input_dir").short('f').long("file"))
+        .arg(Arg::new("language").short('l').long("lang").required(true))
+        .arg(Arg::new("pot_title").short('t').long("title"))
+        .arg(Arg::new("output_dir").short('o').long("out"))
+        .get_matches();
+
+    let (input, content) =
+        ingest_input(matches.get_one::<String>("input_dir").map(|d| d.as_str()))?;
```
mgeisler, line 82: I don't know Clap well, but I think this can be done more directly: by using the `derive` feature, we can work with the command line arguments as a typed struct: https://docs.rs/clap/latest/clap/#example.

I think that will be nicer: the `String` and `as_str()` calls here can be removed.

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+    let lang_option = matches.get_one::<String>("language").unwrap().as_str();
```
mgeisler, line 83: Can we make `--lang` optional? Simply use an empty string if it is not given. Then people can add a language by hand later in the generated POT file (if they like).

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+    Ok(output_path)
+}
+
+fn main() -> anyhow::Result<()> {
+    let matches = Command::new("markdown-xgettext")
+        .about("binary that extracts translatable text from markdown file for mdbook-i18n-helpers")
+        .arg(Arg::new("input_dir").short('f').long("file"))
+        .arg(Arg::new("language").short('l').long("lang").required(true))
+        .arg(Arg::new("pot_title").short('t').long("title"))
+        .arg(Arg::new("output_dir").short('o').long("out"))
```
mgeisler, line 78: Please align the options with `xgettext` from the Gettext utilities. We should only use a few options, I think something like

- `-o`/`--output`: output file, if not given, we use `stdout`.
- `inputfile`: this is not a flag (no `-` or `--`), this is simply the positional arguennts given after the flags. If none are given, we use `stdin`.

Actually, it seems like `xgettext` does not set a `title` or a `lang` in the header entry (see [Filling in the Header Entry](https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html)). So I would suggest leaving them blank as well: less code for us to maintain! :smile:

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    output_path.push(pot_title);
+
+    Ok(output_path)
+}
+
+fn main() -> anyhow::Result<()> {
+    let matches = Command::new("markdown-xgettext")
+        .about("binary that extracts translatable text from markdown file for mdbook-i18n-helpers")
```
mgeisler, line 74:
```suggestion
        .about("Extracts translatable text from Markdown files.")
```

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+    } else {
+        let mut buffer = String::new();
+        io::stdin()
+            .read_to_string(&mut buffer)
+            .context("Failed to read from stdin")?;
+        Ok(("-", buffer))
+    }
+}
+
+fn build_path(output_dir: Option<&str>, pot_title: &str) -> anyhow::Result<PathBuf> {
```
mgeisler, line 53: I would like this function to be simpler:

- if `output` is `"-"`, then write to stdout
- otherwise write to the path given as output. People can create the necessary directories themselves.

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,250 @@
...
+    let output_path = output_path_option.unwrap_or("translated_md_files");
+    let path = PathBuf::from(output_path);
+    fs::create_dir_all(&path)
+        .with_context(|| format!("Failed to create directory: {}", path.display()))?;
+    Ok(path)
+}
+
+fn main() -> anyhow::Result<()> {
+    let matches = Command::new("markdown-gettext")
+        .about("markdown translator binary for mdbook-i18n-helpers")
```
mgeisler, line 88:
```suggestion
        .about("Translate Markdown files using a Gettext PO file.")
```

i18n-helpers/src/bin/markdown-gettext.rs
```diff
@@ -0,0 +1,250 @@
...
+
+fn build_output_path(output_path_option: Option<&str>) -> anyhow::Result<PathBuf> {
+    let output_path = output_path_option.unwrap_or("translated_md_files");
+    let path = PathBuf::from(output_path);
+    fs::create_dir_all(&path)
+        .with_context(|| format!("Failed to create directory: {}", path.display()))?;
+    Ok(path)
+}
+
+fn main() -> anyhow::Result<()> {
```
mgeisler, line 86: I'm a bit in doubt about the best command line interface for this...

One idea would be to allow

```shell
markdown-gettext xx.po foo.md -o foo-xx.md
```

which would translate `foo.md` using `xx.po` and write the output to `foo-xx.md`.

Here, `foo.md` could be left out or it could be `-` and then we read from `stdin` (like for `markdown-xgettext` below). Similar, `-o` can be left out or be `-` and then we write to `stdout`.

If there are multiple inputs, then I guess the user cannot write to `stdout` and must instead give an output directory:

```shell
markdown-gettext xx.po foo.md bar.md --output-dir xx
```

In this case it would be an error to attempt to read from `stdin` since we don't really know what to name the output file in that case.

friendlymatthew: If I'm understanding correctly, we want optional flags for output file title and directory. 

Since `markdown-gettext xx.po foo.md -o foo-xx.md`, does this mean if one was to add an additional wef.md, would the command be: `markdown-gettext xx.po foo.md -o foo-xx.md wef-xx.md`?

Output directory makes sense to me 👍. I would just like some clarification on file naming conventions, especially when one wants to translate multiple files.

mgeisler: > If I'm understanding correctly, we want optional flags for output file title and directory.

Yeah, I'm trying to figure out how we can model the behavior here to be similar to existing Unix commands.

We should also keep it simple! Less code for all of us to write and maintain :smile: 

I'm thinking of using [`cp`](https://man7.org/linux/man-pages/man1/cp.1.html) as a basis:

- `cp` can copy a single path to a new name _or_
- `cp` can copy multiple path into a single directory

The paths which are copied can be files or directories (or a mix of both). If there is a directory, then you must add `-r` to do a recursive copy.

I now see

```
markdown-gettext xx.po source.md dest.md
```
as

```
cp source.md dest.md
```
but with the translation happening as well.

Similarly, 

```
markdown-gettext xx.po foo.md bar.md dest/
```
as

```
cp foo.md bar.md dest/
```

Finally, the rule out `-r` should be implemented: if you want to recursively translate a directory, you have to add a `-r` flag.
 
This new thinking is different from what I had before where I thought we should use the flags from the GNU Gettext utilities. Back then, I hadn't realized that there isn't a command line program which mirrors what our `markdown-gettext` program will do :smile:

Note that these are just ideas, please let me know if you have better ones!

friendlymatthew: @mgeisler I'll write up a doc that outlines what the CLI should do for both `xgettext` and `gettext`. Once we get everything to your liking, I'll change the binaries to follow those requirements!

mgeisler: Sounds great! That document could simultaneously become the manual for the tools, so your work here will be reused right away.

i18n-helpers/src/bin/markdown-xgettext.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    for (lineno, msgid) in extract_messages(content) {
+        let source = format!("{}:{}", source_path, lineno);
+        add_message(&mut catalog, &msgid, &source);
+    }
+
+    Ok(catalog)
+}
+
+fn ingest_input(input_dir: Option<&str>) -> anyhow::Result<(&str, String)> {
```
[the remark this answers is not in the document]
friendlymatthew: Here's the writeup for `xgettext`. (I'll put this in a readme once I get the green light @mgeisler)

# `markdown-gettext`
This tool extracts translatable text from a given Markdown file or stdin and outputs it to a .po file or stdout.

### Usage
> md-xgettext [inputfile] -o/--output

#### Arguments
1. `inputfile`
   - a specified file path
   - if not provided, the program reads from stdin

#### Options
1. `-o/--output`
   - specifies the output file name
   - if not provided, the program writes to stdout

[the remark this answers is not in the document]
friendlymatthew: @mgeisler do you like this? Happy to implement once I get your approval :)

i18n-helpers/src/bin/markdown-gettext.rs
mgeisler: Hey @friendlymatthew, sorry for the slow reply. The above looks great, thank you very much for writing it up! I think it follow the behavior from the [`cp` manual page](https://man7.org/linux/man-pages/man1/cp.1.html) closely.

i18n-helpers/src/bin/markdown-xgettext.rs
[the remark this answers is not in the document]
mgeisler: This seems like a great start, thanks!

Looking at this again, I think we'll eventually want to shorten and simplify the names here. I'm okay with `mdbook-xgettext` and `mdbook-gettext` as plugins for `mdbook`, but these programs will be run by users on a command line. So they should probably be shorter and more clear. Perhaps `md-extract` and `md-translate`? We can move things around after we merge the PR, so don't worry about this too much right now.

i18n-helpers/src/bin/markdown-gettext.rs
friendlymatthew: No worries @mgeisler, I appreciate the opportunity to work on this 😁. 

When I have some time off from official work, I'll get on this ASAP

i18n-helpers/src/bin/markdown-xgettext.rs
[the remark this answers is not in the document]
friendlymatthew: Makes sense -- will update my files accordingly

PR #96, opened 2023-10-04
Fix caching of `target` dir for `fuzz` project

kdarkhan:
Currently caching does not work for `fuzz` job which can be verified by looking at Github workflow outputs.

The main cause of the issue is the config `workspaces: fuzz -> target`. The fuzz target does not use a proper workspace but a separate crate. There is [a pending issue](https://github.com/rust-fuzz/cargo-fuzz/issues/345) discussing that.

The `rust-cache` action is executed in the root of the project and is not able to locate `fuzz` directory by workspace name.

I decided to list the directories to cache instead of relying on target dir detection which seems to work.

I also fixed the path `fuzz/corpus` to `i18n-helpers/fuzz/corpus`.

I adjusted the caching condition `save-if` to cache only on `main` builds. I think this makes sense because we don't need to cache PRs which might not get merged or get updated multiple times during review.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ jobs:
         uses: actions/checkout@v3
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Test
         run: cargo test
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Thanks, this looks good. Before, I let the Cargo-aware caching handle the `fuzz` directory and used the regular caching system for the plain text fuzz corpus. Using the Cargo caching for both ought to work as well.

mgeisler:
@sakex, I believe we forgot this in #86.

sakex:
Oh sorry I missed this and thanks for the fix

PR #100, opened 2023-10-18
Allow formatting in the `SUMMARY.md` file

### LINKED ISSUE #66: Lift restriction on having no formatting in `SUMMARY.md`
Right now, `mdbook-xgettext` forbids formatting in the `SUMMARY.md` file. It does this because the formatting is stripped by `mdbook`, meaning that the plugin receives the chapter name without formatting and so we cannot find the chapter name with a literal search in `SUMMARY.md`.

We should remove this restriction since it makes `mdbook-xgettext` incompatible with `mdbook` projects in the wild, such as the [Rust Book](https://github.com/rust-lang/book/blob/main/src/SUMMARY.md?plain=1#L35-L36).

We should instead do something like this

- Fall back on an inaccurate line number for chapter titles which we cannot find (easy),
- Parse `SUMMARY.md` ourselves, strip the formatting, and do the comparison this way (slightly harder, but we can reuse code from `mdbook`).

mgeisler:
This first changes `mdbook-xgettext` so that formatting is preserved from the `SUMMARY.md` file.

Next, `mdbook-gettext` is changed to translate text from the `SUMMARY.md` file without formatting: this is necessary since `mdbook` strips formatting before sending the TOC to the preprocessor.

Fixes #66.

### DIFF
--- a/i18n-helpers/src/bin/mdbook-gettext.rs
+++ b/i18n-helpers/src/bin/mdbook-gettext.rs
@@ -30,9 +30,28 @@ use mdbook::preprocess::{CmdPreprocessor, PreprocessorContext};
 use mdbook::BookItem;
 use mdbook_i18n_helpers::{extract_events, reconstruct_markdown, translate_events};
 use polib::catalog::Catalog;
+use polib::message::Message;
 use polib::po_file;
+use pulldown_cmark::Event;
 use semver::{Version, VersionReq};
 use std::{io, process};
+
+/// Strip formatting from a Markdown string.
+///
+/// The string can only contain inline text. Formatting such as
+/// emphasis and strong emphasis is removed.
+///
+/// Modelled after `mdbook::summary::stringify_events`.
+fn strip_formatting(text: &str) -> String {
+    extract_events(text, None)
+        .iter()
+        .filter_map(|(_, event)| match event {
+            Event::Text(text) | Event::Code(text) => Some(text.as_ref()),
+            Event::SoftBreak => Some(" "),
+            _ => None,
+        })
+        .collect()
+}
 
 fn translate(text: &str, catalog: &Catalog) -> String {
     let events = extract_events(text, None);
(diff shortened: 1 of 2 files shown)


mgeisler:
TODO: I would like to add an end-to-end test of this: something which verifies that what we extract in `mdbook-xgettext` works with what `mdbook-gettext` expects.

mgeisler:
> TODO: I would like to add an end-to-end test of this: something which verifies that what we extract in `mdbook-xgettext` works with what `mdbook-gettext` expects.

I suppose the unit tests here are good enough — and we'll be testing this soon enough in Comprehensive Rust.

mgeisler:
@dyoo, does this look reasonable to you?

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -30,17 +30,64 @@ use mdbook::preprocess::{CmdPreprocessor, PreprocessorContext};
...
+/// While it is permissible to include formatting in the `SUMMARY.md`
+/// file, `mdbook` will strip it out when rendering the book. It will
+/// also strip formatting when sending the book to preprocessors.
+///
+/// To be able to find the translations for the `SUMMARY.md` file, we
+/// append versions of these messages stripped of formatting.
+fn add_stripped_summary_translations(catalog: &mut Catalog) {
+    let mut stripped_messages = Vec::new();
+    for msg in catalog.messages() {
+        if !msg.source().contains("SUMMARY.md") {
```
dyoo, line 74: Perhaps `ends_width` instead of `contains`?

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -30,17 +30,64 @@ use mdbook::preprocess::{CmdPreprocessor, PreprocessorContext};
...
+        }
+
+        let message = Message::build_singular()
+            .with_msgid(strip_formatting(msg.msgid()))
+            .with_msgstr(strip_formatting(msg.msgstr().unwrap()))
+            .done();
+        stripped_messages.push(message);
+    }
+
+    for msg in stripped_messages.into_iter() {
```
dyoo, line 85: Can we omit "into_iter()"?  I thought Rust would automatically apply this on the thing IntoIterable that the for loop is working over.

dyoo (approved):
LGTM!

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: Ah, that would be nice, but we're actually matching against the full list of filenames here: something like `src/foo/bar.md:123 src/SUMMARY.md:234`.

I could parse this and it would look something like this:

```rust
        if !msg
            .source()
            .split_whitespace()
            .filter_map(|source| source.split_once(':'))
            .map(|(path, _)| Path::new(path))
            .any(|path| path.file_name() == Some(OsStr::new("SUMMARY.md")))
        {
            continue;
        }
```

However, now that I just tried this out, I realize that we cannot make any strong assumptions on where the `SUMMARY.md` file is: `mdbook` allows you to configure the [source directory](https://rust-lang.github.io/mdBook/format/configuration/general.html#general-metadata).

So the code above is essentially the same as just checking if `SUMMARY.md` is found anywhere in the string.

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: Nice, you're spot on!

PR #104, opened 2023-10-20
Define a fuzzer for gettext

kdarkhan:


i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -25,27 +25,20 @@
...
-    let translated_events = translate_events(&events, catalog);
-    let (translated, _) = reconstruct_markdown(&translated_events, None);
-    translated
-}
+use mdbook_i18n_helpers::gettext::translate_book;
+use polib::po_file;
+use std::process;
 
-fn translate_book(ctx: &PreprocessorContext, mut book: Book) -> anyhow::Result<Book> {
+fn parse_catalog(ctx: &PreprocessorContext) -> anyhow::Result<Option<Catalog>> {
```
mgeisler, line 37: The return type here is a bit odd to me: we can successfully parse a `None` — which seems to mean that we decided successfully decided that we don't need to parse the catalog.

Could the function be split up into a few parts? Something like this:

- `fn should_translate(...) -> bool`
- `fn load_catalog(...) -> anyhow::Result<Catalog>`
- `fn translate_book(..., catalog: &Catalog) -> Book`

Then you would only attempt to do I/O and load the catalog if it was determined that we need to do a tranlation — and you would only translate the book if you successfully loaded the catalog.

mgeisler: Oh, and now that I'm in asking mode: could we get docstrings for the functions — maybe even for the ones you move around? Pretty-please? :slightly_smiling_face:

i18n-helpers/fuzz/src/lib.rs
mgeisler, line 1: It's great that you create a little library for this :+1:

mgeisler: and especially nice that the library can live here so that it doesn't impact the rest of the code.

i18n-helpers/fuzz/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    catalog
+}
+
+/// Generate a random Book for fuzzing.
+///
+/// The first u32 value in the tuple determines the type
+/// of a BookItem to insert into the book. The generated Book
+/// is not a fully valid book and contains only fields
+/// which are important for fuzzing.
+pub fn create_book(book_items: Vec<(u32, &str, &str)>) -> Book {
```
mgeisler, line 27: This function can totally work, but it might be interesting for you to play with implementing `Arbitrary` for the `BookItem` type, see the [fuzzing documentation](https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html).

Since we don't own the `BookItem` type, you will need to create a small wrapper type and then generate an arbitrary instance of this — and finally implement [`Into<mdbook::BookItem> for BookItem`](https://doc.rust-lang.org/stable/std/convert/trait.Into.html) where the second `BookItem` is your wrapper or shadow type.

i18n-helpers/src/bin/mdbook-gettext.rs
kdarkhan: Yeah, I agree. I refactored this a bit and decided to make `fn should_translate(ctx: &PreprocessorContext) -> anyhow::Result<bool>`. Others are OK as you suggested. Let me know what you think.

i18n-helpers/fuzz/src/lib.rs
kdarkhan: Thanks for suggestion, using `Into` in conjunction with `Arbitrary` macro makes it easier to read.

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -25,111 +25,61 @@
...
+    if ctx.config.book.language.is_none() {
+        return Ok(false);
     }
 
-    for msg in stripped_messages {
-        catalog.append_or_update(msg);
-    }
+    let path = get_catalog_path(ctx)?;
+    // Nothing to do if PO file is missing.
+    return Ok(path.try_exists()?);
```
mgeisler, line 49: I think this is the same as `path.try_exists()` — perhaps with a `.into()` or so tacked on?

You should at least trip the `return` which is not necessary here.

mgeisler: Ah, looking again, I think we could get away with letting this function return `bool` directly — and then you use a [`is_ok`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.is_ok) to convert the `Result` here. That simplified the other functions a tiny bit too.

kdarkhan: Sounds good, I flattened the type and now it returns just `bool`.

i18n-helpers/fuzz/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+            .done();
+        catalog.append_or_update(message);
+    }
+    catalog
+}
+
+/// Generate a random Book for fuzzing.
+pub fn create_book(book_items: Vec<MyBookItem>) -> Book {
+    let mut book = Book::new();
+    book_items.into_iter().for_each(|item| {
```
mgeisler, line 25: Please use a normal `for` loop — I don't really know why one would ever use `for_each` over a `for` loop.

i18n-helpers/fuzz/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        book.push_item(item);
+    });
+    book
+}
+
+/// Wrapper enum for generating arbitrary `BookItem`s.
+#[derive(Arbitrary, Debug)]
+pub enum MyBookItem {
```
mgeisler, line 33: Please call it `BookItem`, I don't use the `My` prefix much outside of personal toy projects.

Then the `From` impl below will end up looking like

```rust
impl From<BookItem> for mdbook::book::BookItem
```

which nicely shows how the local type (unqualified) relates to the foreign type.

kdarkhan: Done.

i18n-helpers/fuzz/src/lib.rs
kdarkhan: Done.

mgeisler (approved):
Thanks a lot!

kdarkhan:
Can this be merged now? :)

mgeisler:
> Can this be merged now? :)

Yes, I don't know why I didn't click the button right away :smile:

PR #106, opened 2023-10-23
Add translator comment parsing

dyoo:
This moves the comment directive parsing to a separate module and adds two features:

* Customizable prefix for the directive.   It defaults to "i18", but to maintain compatibility, overridden with "mdbook-xgettext"  in its current usage.
* Can parse translator comments.

This is preparatory work to thread this state during message extraction.

### DIFF
--- a/i18n-helpers/src/directives.rs
+++ b/i18n-helpers/src/directives.rs
@@ -1,0 +1,104 @@
+use regex::Regex;
+use std::sync::OnceLock;
+
+#[derive(Debug, PartialEq)]
+pub enum Directive {
+    Skip,
+    TranslatorComment(String),
+}
+
+pub fn find(html: &str) -> Option<Directive> {
+    static RE: OnceLock<Regex> = OnceLock::new();
+    let re = RE.get_or_init(|| {
+        let pattern = r"(?x)
+              <!-{2,}\s*                  # the opening of the comment
+              (?:i18n|mdbook-xgettext)    # the prefix
+              \s*:                        # delimit between prefix and command
+              (?<command>.*[^-])          # the command part of the prefix
+              -{2,}>                      # the closing of the comment
+        ";
+        Regex::new(pattern).expect("well-formed regex")
+    });
+
+    let captures = re.captures(html.trim())?;
+
+    let command = captures["command"].trim();
+    match command.split(is_delimiter).next() {
+        Some("skip") => Some(Directive::Skip),
+        Some("comment") => {
+            let start_of_comment_offset = std::cmp::min(
+                command.find("comment").unwrap() + "comment".len() + 1,
+                command.len(),
+            );
+            Some(Directive::TranslatorComment(
+                command[start_of_comment_offset..].trim().into(),
+            ))
+        }
+        _ => None,
+    }
+}
+
+fn is_delimiter(c: char) -> bool {
+    c.is_whitespace() || c == ':' || c == '-'
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_is_comment_skip_directive_simple() {
+        assert!(matches!(find("<!-- i18n:skip -->"), Some(Directive::Skip)));
+    }
+
+    #[test]
+    fn test_is_comment_skip_directive_tolerates_spaces() {
+        assert!(matches!(find("<!-- i18n: skip -->"), Some(Directive::Skip)));
+    }
+
+    #[test]
+    fn test_is_comment_skip_directive_tolerates_dashes() {
+        assert!(matches!(
+            find("<!--- i18n:skip ---->"),
+            Some(Directive::Skip)
+        ));
+    }
+
+    #[test]
+    fn test_is_comment_skip_directive_needs_skip() {
+        assert!(find("<!-- i18n: foo -->").is_none());
+    }
+
+    #[test]
+    fn test_is_comment_skip_directive_needs_to_be_a_comment() {
+        assert!(find("<div>i18: skip</div>").is_none());
+    }
+
+    #[test]
+    fn test_different_prefix() {
+        assert!(matches!(
+            find("<!-- mdbook-xgettext:skip -->"),
+            Some(Directive::Skip)
+        ));
+    }
+
+    #[test]
+    fn test_translator_comment() {
+        assert!(match find("<!-- i18n:comment: hello world! -->") {
+            Some(Directive::TranslatorComment(s)) => {
+                s == "hello world!"
+            }
+            _ => false,
+        });
+    }
+
+    #[test]
+    fn test_translator_empty_comment_does_nothing() {
+        assert!(match find("<!-- i18n:comment -->") {
+            Some(Directive::TranslatorComment(s)) => {
+                s.is_empty()
+            }
+            _ => false,
+        });
+    }
+}
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -26,11 +26,11 @@
 use polib::catalog::Catalog;
 use pulldown_cmark::{CodeBlockKind, Event, LinkType, Tag};
 use pulldown_cmark_to_cmark::{cmark_resume_with_options, Options, State};
-use regex::Regex;
 use std::sync::OnceLock;
 use syntect::easy::ScopeRangeIterator;
 use syntect::parsing::{ParseState, Scope, ScopeStack, SyntaxSet};
 
+pub mod directives;
 pub mod gettext;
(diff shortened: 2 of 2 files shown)


i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl DirectiveFinder {
+    pub fn new(prefix: &str) -> Self {
+        let pattern: String = r"<!-{2,}\s*".to_owned()
+            + &escape(prefix)
+            + &r"\s*[:-]"
+            + &r"(.*[^-])"  // the command part of the prefix, which must not end in a hyphen.
+            + &r"-{2,}>";
```
mgeisler, line 25: Please put this into a single `format!` call instead. That should cause fewer allocations and also make it clearer that this is one big string. You might want to use the ["verbose" mode](https://docs.rs/regex/latest/regex/#example-verbose-mode) if that makes things clearer.

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
+use regex::{escape, Regex};
+
+#[derive(Debug, PartialEq)]
+pub enum Directive {
+    Skip,
+    TranslatorComment(String),
+}
+
+pub struct DirectiveFinder {
+    pattern_re: Regex,
```
mgeisler, line 10: I would suggest either `pattern` or just `re` — the current name seems a bit long and it's almost like you're encoding the type (`Regex`) into it (`_re`).

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
...
+        Self {
+            pattern_re: Regex::new(&pattern).expect("well-formed regex"),
+        }
+    }
+
+    pub fn find(&self, html: &str) -> Option<Directive> {
+        let captures = self.pattern_re.captures(html.trim())?;
+
+        let command = captures.get(1)?.as_str().trim();
+        match command.split(is_delimiter).next() {
```
mgeisler, line 35: It feels like this could be handled by a (named) group in the regular expression?

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn default() -> Self {
+        Self::new("i18")
+    }
+}
+
+impl DirectiveFinder {
+    pub fn new(prefix: &str) -> Self {
+        let pattern: String = r"<!-{2,}\s*".to_owned()
+            + &escape(prefix)
+            + &r"\s*[:-]"
```
mgeisler, line 23: What does the `-` in the character class match?

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    pub fn find(&self, html: &str) -> Option<Directive> {
+        let captures = self.pattern_re.captures(html.trim())?;
+
+        let command = captures.get(1)?.as_str().trim();
+        match command.split(is_delimiter).next() {
+            Some("skip") => Some(Directive::Skip),
+            Some("comment") => {
```
mgeisler, line 37: I think the regular expression could be made simpler if you write it out in full like this:

```
Regex::new(r"<!-{2,}\s*(i18n|mdbook-xgettext)\s*:\s*(skip|comment)\s*-{2,}>")
```

That is, hard-code the two options (they're linked directly to the enum) and also just hard-code the available prefixes.

I suggest that we clean up the mdbook-xgettext one as soon as we can and that simplifies things a bit more. In fact, we can clean it up right away: when we upgrade to the new version of `mdbook-i18n-helpers`, then we simply search-replace all of Comprehensive Rust in one atomic operation.

A small [search on GitHub](https://github.com/search?q=mdbook-xgettext+skip+path%3A*.md+NOT+repo%3Agoogle%2Fcomprehensive-rust+&type=code) shows that we're likely the only ones that use this feature right now, so this is another good reason to clean up any cruft right away.

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,130 @@
+use regex::{escape, Regex};
+
+#[derive(Debug, PartialEq)]
+pub enum Directive {
+    Skip,
+    TranslatorComment(String),
+}
+
+pub struct DirectiveFinder {
```
mgeisler, line 9: With my idea below of hard-coding the regex, you might not need this struct. You can add methods to enums just like with strucst, so I suggest adding a

```rust
pub fn find(html: &str) -> Option<Self>
```

method to `Directive`.

mgeisler:
This needs a little rebase before it can be tested and merged.

i18n-helpers/src/directives.rs
dyoo: I wanted to support either `<!-- i18-comment: ... -->` or `<!-- i18:comment: ... -->`; perhaps this is too much flexibility?   I'm very open to simplify this; what would you like?

mgeisler: Ah, I see! I perfer the simplest solution with less flexibility — over the years, I've seen that less flexibility normally leads to systems that are easier to understand and maintain :smile:

i18n-helpers/src/directives.rs
dyoo: Acknowledged; I will simplify this and just call it re.

i18n-helpers/src/directives.rs
dyoo: Sounds good: switching to a named group.

i18n-helpers/src/directives.rs
dyoo: Sounds good; simplifying so it just uses `:` between the "i18n" and the command portion.

i18n-helpers/src/directives.rs
dyoo: Acknowledged; I've applied the suggested hardcodings to simplify a bit.  I do think I want to keep the verbose regex (spread across a few lines with comments) because I'm finding it easier to read, especially because there's some subtlety toward the end of the regex.

i18n-helpers/src/directives.rs
dyoo: Acknowledged.  Demoted the method to a plain function in the `directives` module.

dyoo:
Rebased and applied recommended cleanups; once it looks good to you, I'll squash the fixup commits into the branch.

i18n-helpers/src/directives.rs
mgeisler: > I do think I want to keep the verbose regex (spread across a few lines with comments) because I'm finding it easier to read, especially because there's some subtlety toward the end of the regex.

Yeah, I'm fully supportive of using the verbose form! It's made exactly for complex cases like this :smile:

i18n-helpers/src/directives.rs
```diff
@@ -0,0 +1,104 @@
...
+        Regex::new(pattern).expect("well-formed regex")
+    });
+
+    let captures = re.captures(html.trim())?;
+
+    let command = captures["command"].trim();
+    match command.split(is_delimiter).next() {
+        Some("skip") => Some(Directive::Skip),
+        Some("comment") => {
+            let start_of_comment_offset = std::cmp::min(
```
mgeisler, line 29: Oh, I finally understand why we have this complexity: we're accepting things like

```
<!-- i18n: skip -->
<!-- i18n: comment foo bar -->
```

where you parse the first word of the `comment foo bar` part if you see that this is a comment.

If I got that right, then I suggest we change our minds :grimacing: and make this less ambigious. Something like this:

```
<!-- i18n-skip -->
<!-- i18n: foo bar -->
```

I'm thinking of a `function-name: arguments` syntax here. So `i18n-skip` would be a recognized "function" and it takes no arguments. Similarly, `i18n` would be a recognized function and its arguments is a translator comment.

We can merge this already now and then change our minds as we get a bit of experience with this.

mgeisler (approved):
I'm happy to merge this now and then refactor/update as we get more experience.

i18n-helpers/src/directives.rs
dyoo: Agreed; let 's merge first, and then I will revise to use `function-name: arguments`.

dyoo:
Sounds good.   Please merge.

Thanks for your patience as I'm still catching up on responding to your comments.   I'll try to be a bit more prompt in the future.

mgeisler:
> Thanks for your patience as I'm still catching up on responding to your comments. I'll try to be a bit more prompt in the future.

All good! It's your PR and you can drive it at whatever speed suits you :slightly_smiling_face:  This feature is not blocking anything directly, so it's not something I'm trying to push super hard.

PR #107, opened 2023-10-24
Prepare `extract_messages` to return translator comments

dyoo:
Part of #63.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -151,19 +151,17 @@ pub enum Group<'a> {
     ///
     /// This includes `[Text("foo")]` as well as sequences with text
     /// such as `[Start(Emphasis), Text("foo") End(Emphasis)]`.
-    Translate(Vec<(usize, Event<'a>)>, TranslateMetadata),
+    Translate {
+        events: Vec<(usize, Event<'a>)>,
+        /// A comment that may be associated with the translation text.
+        comment: String,
+    },
 
     /// Markdown events which should be skipped when translating.
     ///
     /// This includes structural events such as `Start(Heading(H1,
     /// None, vec![]))`.
     Skip(Vec<(usize, Event<'a>)>),
-}
-
-/// Store some extra information for a translation group.
-#[derive(Debug, Clone, PartialEq, Default)]
-pub struct TranslateMetadata {
-    pub comments: Vec<String>,
 }
 
 #[derive(Debug, Default)]
@@ -190,7 +188,7 @@ impl GroupingContext {
 /// # Examples
 ///
 /// ```
-/// use mdbook_i18n_helpers::{extract_events, group_events, Group, TranslateMetadata};
+/// use mdbook_i18n_helpers::{extract_events, group_events, Group};
 /// use pulldown_cmark::{Event, Tag};
 ///
 /// let events = extract_events("- A list item.", None);
@@ -213,9 +211,10 @@ impl GroupingContext {
 ///             (1, Event::Start(Tag::List(None))),
 ///             (1, Event::Start(Tag::Item)),
 ///         ]),
-///         Group::Translate(vec![
-///             (1, Event::Text("A list item.".into())),
-///         ], TranslateMetadata::default()),
+///         Group::Translate {
+///             events: vec![
+///                 (1, Event::Text("A list item.".into())),
+///             ], comment: "".into()},
 ///         Group::Skip(vec![
 ///             (1, Event::End(Tag::Item)),
 ///             (1, Event::End(Tag::List(None))),
@@ -249,12 +248,10 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
                         parse_codeblock(&events[start..idx], ctx)
                     } else {
                         (
-                            vec![Group::Translate(
-                                events[start..idx].into(),
-                                TranslateMetadata {
-                                    comments: std::mem::take(&mut ctx.comments),
-                                },
-                            )],
+                            vec![Group::Translate {
+                                events: events[start..idx].into(),
+                                comment: std::mem::take(&mut ctx.comments).join(" "),
+                            }],
                             ctx,
                         )
                     }
@@ -375,10 +372,10 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
     }
 
     match state {
-        State::Translate(start) => groups.push(Group::Translate(
-            events[start..].into(),
-            TranslateMetadata::default(),
-        )),
+        State::Translate(start) => groups.push(Group::Translate {
+            events: events[start..].into(),
+            comment: "".into(),
+        }),
         State::Skip(start) => groups.push(Group::Skip(events[start..].into())),
     }
 
@@ -425,12 +422,10 @@ fn heuristic_codeblock<'a>(
 
     if is_translate {
         (
-            vec![Group::Translate(
-                events.into(),
-                TranslateMetadata {
-                    comments: std::mem::take(&mut ctx.comments),
-                },
-            )],
+            vec![Group::Translate {
+                events: events.into(),
+                comment: std::mem::take(&mut ctx.comments).join(" "),
+            }],
             ctx,
         )
     } else {
@@ -469,12 +464,10 @@ fn parse_codeblock<'a>(
 
                 let Ok(ops) = ps.parse_line(text, ss) else {
                     // If parse is failed, the text event should be translated.
-                    ret.push(Group::Translate(
-                        events[idx..idx + 1].into(),
-                        TranslateMetadata {
-                            comments: std::mem::take(&mut ctx.comments),
-                        },
-                    ));
+                    ret.push(Group::Translate {
+                        events: events[idx..idx + 1].into(),
+                        comment: std::mem::take(&mut ctx.comments).join(" "),
+                    });
                     continue;
                 };
 
@@ -513,12 +506,10 @@ fn parse_codeblock<'a>(
                     } else {
                         let whitespace_events = extract_trailing_whitespaces(&mut translate_events);
                         if !translate_events.is_empty() {
-                            groups.push(Group::Translate(
-                                std::mem::take(&mut translate_events),
-                                TranslateMetadata {
-                                    comments: std::mem::take(&mut ctx.comments),
-                                },
-                            ));
+                            groups.push(Group::Translate {
+                                events: std::mem::take(&mut translate_events),
+                                comment: std::mem::take(&mut ctx.comments).join(" "),
+                            });
                         }
                         if !whitespace_events.is_empty() {
                             groups.push(Group::Skip(whitespace_events));
@@ -529,12 +520,10 @@ fn parse_codeblock<'a>(
 
                 let whitespace_events = extract_trailing_whitespaces(&mut translate_events);
                 if !translate_events.is_empty() {
-                    groups.push(Group::Translate(
-                        std::mem::take(&mut translate_events),
-                        TranslateMetadata {
-                            comments: std::mem::take(&mut ctx.comments),
-                        },
-                    ));
+                    groups.push(Group::Translate {
+                        events: std::mem::take(&mut translate_events),
+                        comment: std::mem::take(&mut ctx.comments).join(" "),
+                    });
                 }
                 if !whitespace_events.is_empty() {
                     groups.push(Group::Skip(whitespace_events));
@@ -542,12 +531,10 @@ fn parse_codeblock<'a>(
 
                 if stack_failure {
                     // If stack operation is failed, the text event should be translated.
-                    ret.push(Group::Translate(
-                        events[idx..idx + 1].into(),
-                        TranslateMetadata {
-                            comments: std::mem::take(&mut ctx.comments),
-                        },
-                    ));
+                    ret.push(Group::Translate {
+                        events: events[idx..idx + 1].into(),
+                        comment: std::mem::take(&mut ctx.comments).join(" "),
+                    });
                 } else {
                     ret.append(&mut groups);
                 }
@@ -704,7 +691,7 @@ pub fn extract_messages(document: &str) -> Vec<(usize, ExtractedMessage)> {
 
     for group in group_events(&events) {
         match group {
-            Group::Translate(events, group_extra) => {
+            Group::Translate { events, comment } => {
                 if let Some((lineno, _)) = events.first() {
                     let (text, new_state) = reconstruct_markdown(&events, state);
                     // Skip empty messages since they are special:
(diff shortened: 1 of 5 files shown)


i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -93,9 +94,15 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
...
-            for (lineno, msgid) in extract_messages(&chapter.content) {
+            for (lineno, extracted) in extract_messages(&chapter.content) {
+                let msgid = extracted.message;
                 let source = format!("{}:{}", path.display(), lineno);
-                add_message(&mut catalog, &msgid, &source);
+                add_message(
+                    &mut catalog,
+                    &msgid,
+                    &source,
+                    &extracted.translator_comment.unwrap_or(String::from("")),
```
mgeisler, line 104: Try using

```rust
extracted.translator_comment.as_deref().unwrap_or_default()
```

instead. [Option::as_deref](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.as_deref) is one of those weird methods which you only see in Rust, it turns `Option<String>` into `Option<&str>`, after which `unwrap_or_default()` can give you the string or `""`.

In this case it'll probably generate the same code: `String::from("")` is the same as `String::new()` (I'm not good at reading assembly, but a quick check with [Godbolt](https://rust.godbolt.org/z/db5o8MasG) shows this) and `String::new()` does not allocate.

i18n-helpers/src/lib.rs
```diff
@@ -126,19 +126,32 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'static>>) -> Vec<(
...
+pub struct GroupExtra {
+    translator_comments: Vec<String>,
+}
+
+impl Default for GroupExtra {
+    fn default() -> GroupExtra {
+        GroupExtra {
+            translator_comments: vec![],
+        }
+    }
```
mgeisler, line 154: This looks like the same as what `#[derive(Default)]` on `GroupExtra` would give you.

i18n-helpers/src/lib.rs
```diff
@@ -172,14 +185,14 @@ pub enum Group<'a> {
 ///         Group::Skip(&[
 ///             (1, Event::Start(Tag::List(None))),
 ///             (1, Event::Start(Tag::Item)),
-///         ]),
+///         ], Default::default()),
```
mgeisler, line 188: I much prefer `GroupExtra::default()` over `Default::default()` — using the final type instead of the trait shows me what I'm getting.

i18n-helpers/src/lib.rs
```diff
@@ -219,23 +239,53 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
                 State::Translate(start) => {
                     if ctx.skip_next_group {
                         (
-                            Group::Skip(&events[start..idx]),
+                            Group::Skip(
+                                &events[start..idx],
+                                GroupExtra {
+                                    translator_comments: vec![],
+                                },
```
mgeisler, line 246: This should be the same as:
```rust
GroupExtra::default()
```

i18n-helpers/src/lib.rs
```diff
@@ -219,23 +239,53 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
...
                     } else {
-                        (Group::Translate(&events[start..idx]), ctx)
+                        (
+                            Group::Translate(
+                                &events[start..idx],
+                                GroupExtra {
+                                    translator_comments: ctx.translator_comments.clone(),
+                                },
+                            ),
+                            ctx.clear_translator_comments(),
```
mgeisler, line 268: There is an interesting pattern here: you clone `ctx.translator_comments` and then you clear them right after. So you could skip the clone if you moved the comments instead.

You can do this with a `std::mem::swap`: it can move values between two places without deallocating either.

i18n-helpers/src/lib.rs
```diff
@@ -408,6 +476,20 @@ pub fn reconstruct_markdown(
     (String::from(markdown.trim_matches('\n')), new_state)
 }
 
+#[derive(Debug, PartialEq)]
+pub struct ExtractedMessage {
```
mgeisler, line 480: I'm wondering if you could use [polib](https://docs.rs/polib/latest/polib/index.html)::[message](https://docs.rs/polib/latest/polib/message/index.html)::[Message](https://docs.rs/polib/latest/polib/message/struct.Message.html) directly here?

Basically, let `extract_messages` return partially-filled `Message` structs. Then you would be sure to have all necessary fields right where you need them.

dyoo: The main reason I wanted a separate struct that we own is that I wanted to prepare for other work, for the refactoring of comment skipping as a phase after extract_messages.

With `ExtractedMessage`, it would be easier to add more state that could be used for post-extract skipping.  Using `polib::message::Message` as the return type seemed to interfere with that idea.

Unsure.  Since I haven't yet done the skipping refactoring work, I wouldn't be opposed to following the suggestion.  But alternatively, I could put a `polib::message::Message` within an `ExtractedMessage`.  Would that be another way to simplify that seems okay with you?

mgeisler: > The main reason I wanted a separate struct that we own is that I wanted to prepare for other work, for the refactoring of comment skipping as a phase after extract_messages.

That sounds like a good reason to me — I'm happy to keep our own type then!

mgeisler:
This needs a little rebase before it can be tested and merged.

dyoo:
Sounds good!  I'll sync soon; I plan to do a rebase on top of dalance's work soon, since it looks like they're further along.

i18n-helpers/src/bin/mdbook-xgettext.rs
mgeisler: Put differently, `&... String::from("...")` looks strange: you're constructing a `String` onto to turn it into a `&str` on the same line. That's what made me question this little line :slightly_smiling_face:

dyoo: Sounds good; switched to get_deref().

i18n-helpers/src/lib.rs
dyoo: Acknowledged: switched to derived impl.

i18n-helpers/src/lib.rs
dyoo: Agreed!  Okay, switched.

i18n-helpers/src/lib.rs
dyoo: Done!

i18n-helpers/src/lib.rs
dyoo: Acknowledged; replaced with use of std::mem::take().

i18n-helpers/src/lib.rs
```diff
@@ -1666,4 +1689,27 @@ after-no-comment
...
+        assert_eq!(
+            extract_messages(
+                r#"
+<!-- mdbook-xgettext:comment: greetings! -->
+```python
+print("Hello world")
+```
+"#
+            )
+            .into_iter()
```
mgeisler, line 1704: Is `extract_messages` not already returning a `Vec`?

i18n-helpers/src/lib.rs
```diff
@@ -603,25 +702,37 @@ pub fn reconstruct_markdown(
...
+                        messages.push((
+                            *lineno,
+                            ExtractedMessage {
+                                message: text,
+                                translator_comment: {
+                                    if group_extra.translator_comments.is_empty() {
+                                        None
+                                    } else {
+                                        Some(group_extra.translator_comments.join(" "))
+                                    }
```
mgeisler, line 727: Consider using the little-known [bool::then](https://doc.rust-lang.org/stable/std/primitive.bool.html#method.then) method here:

```suggestion
                                    group_extra.translator_comments
                                        .is_empty()
                                        .then(|| group_extra.translator_comments.join(" "))
```

Alternatively, change `translator_comment` to `String` and model the absence of one with an empty string. As long as polib skips such a comment, this would make your life easier.

i18n-helpers/src/lib.rs
```diff
@@ -603,25 +702,37 @@ pub fn reconstruct_markdown(
...
+                                },
+                            },
+                        ));
                     }
                     state = Some(new_state);
                 }
             }
-            Group::Skip(events) => {
+            Group::Skip(events, _group_extra) => {
```
mgeisler, line 735: Just use `_` for unnecessary parts in a pattern:

```suggestion
            Group::Skip(events, _) => {
```

i18n-helpers/src/lib.rs
```diff
@@ -148,13 +148,33 @@ pub enum Group<'a> {
     ///
     /// This includes `[Text("foo")]` as well as sequences with text
     /// such as `[Start(Emphasis), Text("foo") End(Emphasis)]`.
-    Translate(Vec<(usize, Event<'a>)>),
+    Translate(Vec<(usize, Event<'a>)>, GroupExtra),
```
mgeisler, line 151: Since both enum variants have the same fields, but only one of them seems to need the `GroupExtra` data, I think we could refactor `Group` into a struct in the future.

Something like

```rust
enum Kind { Translate, Skip }

struct Extra {
   pub translator_comments: Vec<String>,
}

struct Group {
    kind: Kind,
    extra: Extra,
}
```

That should give less repetition in the code that uses these values.

You could perhaps just skip the `GroupExtra` field on the `Skip` variant?

i18n-helpers/src/lib.rs
```diff
@@ -315,6 +338,22 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
 
                         ctx.skip_next_group = true;
                     }
+
+                    Some(directives::Directive::TranslatorComment(translator_comment)) => {
```
mgeisler, line 342: We can have a little less "stuttering" by using `comment` to mean `translator_comment` in our local variables:

```suggestion
                    Some(directives::Directive::TranslatorComment(comment)) => {
```

I would also be totally okay with `TranslatorComment` being just `Comment` since I don't think we'll have any other kind of comment?

dyoo: Sounds good!  Dropping the term "translator_comment" everywhere with just "comment".

i18n-helpers/src/lib.rs
dyoo: Let me drop the `GroupExtra` from `Skip` for now; at the moment, we do not need it.  I overgeneralized there.

i18n-helpers/src/lib.rs
dyoo: Okay, done.

i18n-helpers/src/lib.rs
dyoo: Okay, I will simplify to using just a `String` rather than an `Option<String>`; I think I was probably concerned about presence detection for comments when I first wrote this.  But empty string is handled as expected, so I don't need the presence detection after all.

i18n-helpers/src/lib.rs
dyoo: Whooops; thank you.  This must have been some residue from earlier work.  Dropped.

dyoo:
I believe this is close to ready.  Is there anything I've missed before I flatten the history?

dyoo:
My deep apologies for leaving this to wither.  I was sick again for the past week or two, so was unable to work on this for a very long time.

Let me page this back into my memory this week and prepare this for review once more.

i18n-helpers/src/lib.rs
```diff
@@ -160,6 +160,26 @@ pub enum Group<'a> {
     Skip(Vec<(usize, Event<'a>)>),
 }
 
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct Extra {
```
mgeisler, line 164: Please add a doccomment here — the name `Extra` is quite generic, so it'll be hard for a casual reader to know what it's about :sweat_smile:

If the type is produced by a certain function, then it can be useful to link to that function — similarly, if it's supposed to be confused by just one function, then linking to it can be useful too.

dyoo: Hmm... now that I take a fresh look at this, I do not like the extra structure.  Let me absorb its functionality into Group::Translate.  I need to change Group::Translate from a tuple to a struct, but I think it's better than what I was doing before there.

mgeisler: Alright, sounds good to me! Thanks for keeping working on this!

i18n-helpers/src/lib.rs
```diff
@@ -160,6 +164,21 @@ pub enum Group<'a> {
     Skip(Vec<(usize, Event<'a>)>),
 }
 
+#[derive(Debug, Default)]
+struct GroupingContext {
```
mgeisler, line 168: Please add a small docstring here so that people can see what this type is for.

mgeisler: (The type was hidden inside a function before, which makes docstrings less necessary — but now that it's a top-level type, a docstring can be very useful.)

i18n-helpers/src/lib.rs
```diff
@@ -389,14 +421,23 @@ fn heuristic_codeblock<'a>(events: &'a [(usize, Event)]) -> Vec<Group<'a>> {
     };
 
     if is_translate {
-        vec![Group::Translate(events.into())]
+        (
+            vec![Group::Translate {
+                events: events.into(),
+                comment: std::mem::take(&mut ctx.comments).join(" "),
```
mgeisler, line 427: I think this pattern occurs twice: what about making it a method on the `GroupingContext`?

i18n-helpers/src/xgettext.rs
```diff
@@ -39,14 +39,15 @@ fn strip_link(text: &str) -> String {
...
-fn add_message(catalog: &mut Catalog, msgid: &str, source: &str) {
+fn add_message(catalog: &mut Catalog, msgid: &str, source: &str, comment: &str) {
     let sources = match catalog.find_message(None, msgid, None) {
         Some(msg) => format!("{}\n{}", msg.source(), source),
         None => String::from(source),
     };
     let message = Message::build_singular()
         .with_source(sources)
         .with_msgid(String::from(msgid))
+        .with_comments(comment.into())
```
mgeisler, line 50: Let us make this consistent with the line above:

```suggestion
        .with_comments(String::from(comment))
```

Using `.into()` or `T::from()` is a bit of a style choice: I personally like the `T::from()` form since the `T` type is explicit. Others prefer the `.into()` calls instead. Both are valid, we should just be consistent.

mgeisler (approved):
I think this looks very nice! Could you add a small mention of this new feature to the documentation and then merge it? Thanks!

mgeisler:
I merged this now so we can start building on top of it — thanks @dyoo for sticking with it! Perhaps you can send out some small PRs with any follow up changes?

PR #109, opened 2023-10-28
Extract only strings and comments from codeblock

### LINKED ISSUE #95: Extract only strings and comments from code blocks
As a further step after #75, we should offer an option to only extract literal strings and comments from the code.

For [this example](https://google.github.io/comprehensive-rust/basic-syntax/functions-interlude.html):

```rust
fn pick_one<T>(a: T, b: T) -> T {
    if std::process::id() % 2 == 0 { a } else { b }
}

fn main() {
    println!("coin toss: {}", pick_one("heads", "tails"));
    println!("cash prize: {}", pick_one(500, 1000));
}
```

we would end up with just four small strings

- `"coin toss: {}"`
- `"heads"`
- `"tails"`
- `"cash prize: {}"`

in the POT file.

This would require us to process [`Tag::CodeBlock`](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/enum.Tag.html#variant.CodeBlock) in a more fine-grained way, but I think it could be worth it.

The fun part would be to find a cross-language solution. I suspect our best bet would be to use a _syntax highlighting_ library: they normally detect strings and comments and so such a library should have the necessary machinery.

dalance:
Fixes #95 

This PR adds codeblock parsing support to extract only strings and comments.

### DIFF
--- a/i18n-helpers/fuzz/fuzz_targets/group_events.rs
+++ b/i18n-helpers/fuzz/fuzz_targets/group_events.rs
@@ -1,7 +1,7 @@
 #![no_main]
 
 use libfuzzer_sys::fuzz_target;
-use mdbook_i18n_helpers::{extract_events, group_events, Group};
+use mdbook_i18n_helpers::{extract_events, group_events, reconstruct_markdown, Group};
 use pretty_assertions::assert_eq;
 
 fuzz_target!(|text: String| {
@@ -11,8 +11,13 @@ fuzz_target!(|text: String| {
         .flat_map(|group| match group {
             Group::Translate(events) | Group::Skip(events) => events,
         })
-        .cloned()
         .collect::<Vec<_>>();
 
-    assert_eq!(events, flattened_groups);
+    // Comparison through markdown text to detect missing text.
+    // Events can't be compared directly because `group_events`
+    // may split a event into some events.
+    let text_from_events = reconstruct_markdown(&events, None);
+    let text_from_groups = reconstruct_markdown(&flattened_groups, None);
+
+    assert_eq!(text_from_events, text_from_groups);
 });
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -24,10 +24,12 @@
 //! how to use the supplied `mdbook` plugins.
 
 use polib::catalog::Catalog;
-use pulldown_cmark::{Event, LinkType, Tag};
+use pulldown_cmark::{CodeBlockKind, Event, LinkType, Tag};
 use pulldown_cmark_to_cmark::{cmark_resume_with_options, Options, State};
 use regex::Regex;
 use std::sync::OnceLock;
+use syntect::easy::ScopeRangeIterator;
+use syntect::parsing::{ParseState, Scope, ScopeStack, SyntaxSet};
 
 pub mod gettext;
 pub mod normalize;
@@ -128,19 +130,19 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'static>>) -> Vec<(
 }
 
 /// Markdown events grouped by type.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Group<'a> {
     /// Markdown events which should be translated.
     ///
     /// This includes `[Text("foo")]` as well as sequences with text
     /// such as `[Start(Emphasis), Text("foo") End(Emphasis)]`.
-    Translate(&'a [(usize, Event<'a>)]),
+    Translate(Vec<(usize, Event<'a>)>),
 
     /// Markdown events which should be skipped when translating.
     ///
     /// This includes structural events such as `Start(Heading(H1,
     /// None, vec![]))`.
-    Skip(&'a [(usize, Event<'a>)]),
+    Skip(Vec<(usize, Event<'a>)>),
 }
 
 /// Group Markdown events into translatable and skipped events.
@@ -171,14 +173,14 @@ pub enum Group<'a> {
 /// assert_eq!(
 ///     groups,
 ///     vec![
-///         Group::Skip(&[
+///         Group::Skip(vec![
 ///             (1, Event::Start(Tag::List(None))),
 ///             (1, Event::Start(Tag::Item)),
 ///         ]),
-///         Group::Translate(&[
+///         Group::Translate(vec![
 ///             (1, Event::Text("A list item.".into())),
 ///         ]),
-///         Group::Skip(&[
+///         Group::Skip(vec![
 ///             (1, Event::End(Tag::Item)),
 ///             (1, Event::End(Tag::List(None))),
 ///         ]),
@@ -209,27 +211,27 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
     }
 
     impl State {
-        /// Creates a group based on the capturing state and context.
-        fn into_group<'a>(
+        /// Creates groups based on the capturing state and context.
+        fn into_groups<'a>(
             self,
             idx: usize,
             events: &'a [(usize, Event<'a>)],
             ctx: GroupingContext,
-        ) -> (Group<'a>, GroupingContext) {
+        ) -> (Vec<Group<'a>>, GroupingContext) {
             match self {
                 State::Translate(start) => {
                     if ctx.skip_next_group {
                         (
-                            Group::Skip(&events[start..idx]),
+                            vec![Group::Skip(events[start..idx].into())],
                             ctx.clear_skip_next_group(),
                         )
-                    } else if is_nontranslatable_codeblock_group(&events[start..idx]) {
-                        (Group::Skip(&events[start..idx]), ctx)
+                    } else if is_codeblock_group(&events[start..idx]) {
+                        (parse_codeblock(&events[start..idx]), ctx)
                     } else {
-                        (Group::Translate(&events[start..idx]), ctx)
+                        (vec![Group::Translate(events[start..idx].into())], ctx)
                     }
                 }
-                State::Skip(start) => (Group::Skip(&events[start..idx]), ctx),
+                State::Skip(start) => (vec![Group::Skip(events[start..idx].into())], ctx),
             }
         }
     }
@@ -246,18 +248,18 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
             // make the group self-contained.
             Event::Start(Tag::Paragraph | Tag::CodeBlock(..)) => {
                 // A translatable group starts here.
-                let next_group;
-                (next_group, ctx) = state.into_group(idx, events, ctx);
-                groups.push(next_group);
+                let mut next_groups;
+                (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                groups.append(&mut next_groups);
 
                 state = State::Translate(idx);
             }
             Event::End(Tag::Paragraph | Tag::CodeBlock(..)) => {
                 // A translatable group ends after `idx`.
                 let idx = idx + 1;
-                let next_group;
-                (next_group, ctx) = state.into_group(idx, events, ctx);
-                groups.push(next_group);
+                let mut next_groups;
+                (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                groups.append(&mut next_groups);
 
                 state = State::Skip(idx);
             }
@@ -277,9 +279,9 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
                 // If we're currently skipping, then a new
                 // translatable group starts here.
                 if let State::Skip(_) = state {
-                    let next_group;
-                    (next_group, ctx) = state.into_group(idx, events, ctx);
-                    groups.push(next_group);
+                    let mut next_groups;
+                    (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                    groups.append(&mut next_groups);
 
                     state = State::Translate(idx);
                 }
@@ -290,9 +292,9 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
             Event::Html(s) if is_comment_skip_directive(s) => {
                 // If in the middle of translation, finish it.
                 if let State::Translate(_) = state {
-                    let next_group;
-                    (next_group, ctx) = state.into_group(idx, events, ctx);
-                    groups.push(next_group);
+                    let mut next_groups;
+                    (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                    groups.append(&mut next_groups);
 
                     // Restart translation: subtle but should be
                     // needed to handle the skipping of the rest of
@@ -307,9 +309,9 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
             // skipping group.
             _ => {
                 if let State::Translate(_) = state {
-                    let next_group;
-                    (next_group, ctx) = state.into_group(idx, events, ctx);
-                    groups.push(next_group);
+                    let mut next_groups;
+                    (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                    groups.append(&mut next_groups);
 
                     state = State::Skip(idx);
                 }
@@ -318,8 +320,8 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
     }
 
     match state {
-        State::Translate(start) => groups.push(Group::Translate(&events[start..])),
-        State::Skip(start) => groups.push(Group::Skip(&events[start..])),
+        State::Translate(start) => groups.push(Group::Translate(events[start..].into())),
+        State::Skip(start) => groups.push(Group::Skip(events[start..].into())),
     }
 
     groups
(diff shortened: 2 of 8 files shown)


i18n-helpers/src/lib.rs
```diff
@@ -333,17 +360,105 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+                                text_line + text[0..range.start].lines().count() - 1
+                            };
+
+                            let _ = stack.apply_with_hook(scope, |op, scopes| {
+                                let range = range.clone();
+                                if !range.is_empty() {
+                                    let is_translate = match op {
+                                        BasicScopeStackOp::Push(x) => {
+                                            is_translate_scope(x)
+                                                | scopes.iter().any(|x| is_translate_scope(*x))
```
mgeisler, line 417: Should the `|` be `||`? Otherwise there will be no short-circuit here.

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -285,21 +285,20 @@ mod tests {
...
+            ("// line comment\n", "// linie kommentar\n"),
+            ("/* block\ncomment */", "/* block\nkommentar */"),
+        ]);
         assert_eq!(
             translate(
                 "Text before.\n\
                  \n\
                  \n\
                  ```rust,editable\n\
+                 // line comment\n\
```
mgeisler, line 299: Could you also try testing with comments mixed with regular content? Such as

```rust
let x = 42; // line comment

let y = /* before */   false   /* after */;
```

i18n-helpers/src/lib.rs
```diff
@@ -127,19 +129,44 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'static>>) -> Vec<(
...
+#[derive(Debug, Clone, PartialEq)]
 pub enum Group<'a> {
     /// Markdown events which should be translated.
     ///
     /// This includes `[Text("foo")]` as well as sequences with text
     /// such as `[Start(Emphasis), Text("foo") End(Emphasis)]`.
     Translate(&'a [(usize, Event<'a>)]),
 
+    /// Owned version of `Group::Translate`.
+    TranslateOwned(Vec<(usize, Event<'a>)>),
```
mgeisler, line 141: I think it would be okay to remove the borrowing variants and always own the `Event<'a>` values. My thinking is that `mdbook build` is something you run offline, so it's okay if it takes a few 100ms more (especially if it can simplify the code).

i18n-helpers/src/lib.rs
```diff
@@ -333,17 +360,105 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+                _ => {
+                    ret.push(Group::Skip(&events[idx..idx + 1]));
+                }
+            }
         }
-        _ => false,
+        ret
+    } else {
+        // If there is no language specifier, the whole codeblock should be translated.
+        vec![Group::Translate(events)]
```
mgeisler, line 461: I believe this could be an early return at the top of this function?

i18n-helpers/src/lib.rs
```diff
@@ -333,17 +360,105 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+                                    }
+                                }
+                            });
+                        }
+                        if !translate.is_empty() {
+                            ret.push(Group::TranslateOwned(std::mem::take(&mut translate)));
+                        }
+                    } else {
+                        // If parse is failed, the whole codeblock should be translated.
+                        ret.push(Group::Translate(&events[idx..idx + 1]));
```
mgeisler, line 450: This could probably also be moved to the top of the loop — perhaps using a `let-else`: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html.

mgeisler:
Hi @dalance,

> This PR adds codeblock parsing support to extract only strings and comments.

Wow, thank you so much! This looks really good from a quick look!

Could you test how https://google.github.io/comprehensive-rust/zh-CN/hello-world/small-example.html looks after this? In particular, does `mdbook-i18n-normalize` do the right thing here? It should turn [this message](https://github.com/google/comprehensive-rust/blob/main/po/zh-CN.po#L2256-L2288) into 5 individual messages, one for each line comment.

@dyoo has written a lot of this code and I hope he can review this. Also, I think this will collide with #107 and the design he started there to support things like translation comments. So we should make sure we can get everything working nicely together here :slightly_smiling_face: 

I _think_ we would want an option in `book.toml` for this? That could certainly be added later down the road if someone asks for it.

dalance:
Hi @mgeisler,
Thank you for your review.
I fixed some codes which are suggested.

The small-example.md becomes like below.
I think it is an expected result.

```
#: src/hello-world/small-example.md:6
msgid "// Program entry point\n"
msgstr ""

#: src/hello-world/small-example.md:7
msgid "// Mutable variable binding\n"
msgstr ""

#: src/hello-world/small-example.md:8 src/traits/impl-trait.md:15
msgid "\"{x}\""
msgstr ""

#: src/hello-world/small-example.md:8
msgid "// Macro for printing, like printf\n"
msgstr ""

#: src/hello-world/small-example.md:9
msgid "// No parenthesis around expression\n"
msgstr ""

#: src/hello-world/small-example.md:10
msgid "// Math like in other languages\n"
msgstr ""

#: src/hello-world/small-example.md:15
msgid "\" -> {x}\""
msgstr ""
```

The conflicting point with #107 seems to be the modifitation to `Group` enum.
It may be better that splitting this PR to changing to `Group` which owns `Event` and adding `parse_codeblock` function?

> I think we would want an option in book.toml for this? That could certainly be added later down the road if someone asks for it.

This PR changes messages.pot significantly.
So an option to disable this feature may be useful for existing users.

dalance:
I changed some logics to treat continuous line comments.
This is because the previous code can't handle indented continuous line comments.

mgeisler:
Thanks for all the updates!

> > I think we would want an option in book.toml for this? That could certainly be added later down the road if someone asks for it.
> 
> This PR changes messages.pot significantly. So an option to disable this feature may be useful for existing users.

Yeah, I can imagine it will change the files a lot — but for the better!

Please consider adding a line about this new behavior to the documentation. I think it could just be a little paragraph in [Marking Sections to be Skipped for Translation](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/USAGE.md#marking-sections-to-be-skipped-for-translation) which says something like

> Note that we don't extract the full text of code blocks. Only text that is recognized as comments and literal strings is extracted.

Something like that, feel free to rephrase it how you like.

> The small-example.md becomes like below. I think it is an expected result.

Yes, this looks great! Please also try running

```
mdbook-i18n-normalize po/zh-CN.po po/zh-CN-normalized.po
```

and verify that the existing translations are correctly paired up with the new messages. The normalization tool is what is supposed to let us change how we extract messages while also keeping existing translations intact. If I did things correctly, it should "just work", but it'll be nice to have confirmation from you :slightly_smiling_face:

dalance:
The result of mdbook-i18n-normalize is below.
It works fine too.

```
#: src/hello-world/small-example.md:6
msgid "// Program entry point\n"
msgstr "// 程序入口\n"

#: src/hello-world/small-example.md:7
msgid "// Mutable variable binding\n"
msgstr "// 可变变量绑定\n"

#: src/hello-world/small-example.md:8
#: src/traits/impl-trait.md:15
msgid "\"{x}\""
msgstr "\"{x}\""

#: src/hello-world/small-example.md:8
msgid "// Macro for printing, like printf\n"
msgstr "// 与 printf 类似的输出宏\n"

#: src/hello-world/small-example.md:9
msgid "// No parenthesis around expression\n"
msgstr "// 表达式周围没有括号\n"

#: src/hello-world/small-example.md:10
msgid "// Math like in other languages\n"
msgstr "// 与其他语言类似的数值计算\n"

#: src/hello-world/small-example.md:15
msgid "\" -> {x}\""
msgstr "\" -> {x}\""
```

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -285,21 +285,21 @@ mod tests {
...
-            "```rust,editable\n\
-             fn foo() {\n\n    let x = \"hello\";\n\n}\n\
-             ```",
-            "```rust,editable\n\
-             fn FOO() {\n\n    let X = \"guten tag\";\n\n}\n\
-             ```",
-        )]);
+        let catalog = create_catalog(&[
+            ("\"hello\"", "\"guten tag\""),
+            ("// line comment\n", "// linie kommentar\n"),
```
mgeisler, line 290: This is a small thing, but would it be possible to trim away the final whitespace from one or more line comments?

It looks simpler in the PO file afterwards and it's simpler for the translators to deal with. It would also be consistent with how we turn `# Foo` into just `"Foo"` instead of `"Foo\n"`.

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+fn parse_codeblock<'a>(events: &'a [(usize, Event)]) -> Vec<Group<'a>> {
+    // Language detection from language identifier of codeblock.
+    let ss = SyntaxSet::load_defaults_newlines();
+    let syntax = if let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(x)))) = &events[0] {
+        ss.find_syntax_by_token(x.split(',').next().unwrap())
+    } else {
+        None
+    };
+
+    let is_translate_scope = |x| {
```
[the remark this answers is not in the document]
dalance: Thanks.
I fixed it.

i18n-helpers/src/bin/mdbook-gettext.rs
dalance: I agree trimming the final whitespace (and indent spaces) is better for the translators, but I couldn't find an appropriate way for it.

If the linebreak is moved from `Group::Translate` to `Group::Skip`, I thought the continuous line comments are splitted into multiple messages. This causes translation difficulty because multi-lined paragraph can't be translated line by line for the differences of word order and length between languages.

Do you have any idea?

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: Thanks for updating the tests!

i18n-helpers/src/lib.rs
```diff
@@ -24,10 +24,12 @@
 //! how to use the supplied `mdbook` plugins.
 
 use polib::catalog::Catalog;
-use pulldown_cmark::{Event, LinkType, Tag};
+use pulldown_cmark::{CodeBlockKind, CowStr, Event, LinkType, Tag};
 use pulldown_cmark_to_cmark::{cmark_resume_with_options, Options, State};
 use regex::Regex;
 use std::sync::OnceLock;
+use syntect::easy::ScopeRangeIterator;
+use syntect::parsing::{BasicScopeStackOp, ParseState, Scope, ScopeStack, SyntaxSet};
```
mgeisler, line 32: By the way, I would be happy to move this new code block parsing logic to its own little `code_blocks.rs` module (before/after) merging this PR.

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,140 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+            (_, Event::End(Tag::CodeBlock(_)))
+        ]
+    )
+}
+
+/// Returns true if the scope should be translated.
+fn is_translate_scope(x: Scope) -> bool {
+    let scope_string = Scope::new("string").unwrap();
+    let scope_comment = Scope::new("comment").unwrap();
+    scope_string.is_prefix_of(x) | scope_comment.is_prefix_of(x)
```
mgeisler, line 354: I believe it should be fine to only evaluate one side of the `OR` with a Boolean short-circuiting operator:

```suggestion
    scope_string.is_prefix_of(x) || scope_comment.is_prefix_of(x)
```

i18n-helpers/src/lib.rs
```diff
@@ -923,15 +1047,81 @@ The document[^1] text.
...
+            ],
+        );
+    }
+
+    #[test]
+    fn extract_messages_multi_language_code_blocks() {
+        assert_extract_messages(
+            "```c\n\
+            // C\n\
+            'C'; \"C\";\n\
```
mgeisler, line 1104: I think using a raw and un-indented string could be easier to read here.

I'm personally torn between using the `\n\` syntax and raw strings: I like the `\n\` syntax for short strings that don't have leading whitespace. For larger strings, I think we should use raw strings.

i18n-helpers/src/lib.rs
mgeisler: The code moved around and I commented at the new location :smile:

i18n-helpers/src/lib.rs
mgeisler: Thanks for updating!

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: I'll let @dyoo look into the details of the parsing — I think you're doing the right thing by putting multiple consecutive line comments into the same message.

mgeisler:
@henrif75, when this is merged, we will do a 0.3 release. This is a breaking change: we need to run `mdbook-i18n-normalize` on our translations to keep any translated comments intact (as tested by @dalance in https://github.com/google/mdbook-i18n-helpers/pull/109#issuecomment-1784795489). I'll make sure to explain this again in the release notes.

dalance:
I found a normalization issue.
The result of re-executed nomalization to po/zh-CN-normalized.po is here.

```
#: src/hello-world/small-example.md:6
msgid "// Program entry point"
msgstr "// 程序入口"
```

The trailing linebreak is missing.
This is because msgid/msgstr don't have a context which is in codeblock or not.
I can add a workaround like below, but hand-written detection is required because language is not clear.

```rust
pub fn extract_messages(document: &str) -> Vec<(usize, String)> {
    if document.starts_with("//") && document.ends_with("\n") {
        return vec![(1, document.to_string())];
    }
```

How do you think the workaround?
Or any information which shows msgid/msgstr is originated from codeblock should be added to po file?

dalance:
I fixed the reviewed issues.
Additionally, I found my mistake about the usage of syntect stack through fuzz test, and fixed it too.

dalance:
I think adding `codeblock` flag to po file like below to resolve normalization issue may be better.
It seems to work fine at my codebase.

```
#: src/hello-world/small-example.md:6
#, codeblock
msgid "// Program entry point"
msgstr "// 程序入口"
```

henrif75:
> @henrif75, when this is merged, we will do a 0.3 release. This is a breaking change: we need to run `mdbook-i18n-normalize` on our translations to keep any translated comments intact (as tested by @dalance in [#109 (comment)](https://github.com/google/mdbook-i18n-helpers/pull/109#issuecomment-1784795489)). I'll make sure to explain this again in the release notes.

Sounds good. I assume it's going to break ongoing PRs?

dyoo:
I am so sorry for late review comments; I thought I pressed Submit, and it turns out that I did not.  :(

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+    let mut ps = ParseState::new(syntax);
+    let mut ret = vec![];
+
+    for (idx, event) in events.iter().enumerate() {
+        match event {
+            (text_line, Event::Text(text)) => {
+                let mut stack = ScopeStack::new();
+
+                let Ok(ops) = ps.parse_line(text, &ss) else {
+                    // If parse is failed, the whole codeblock should be translated.
```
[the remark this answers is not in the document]
dalance: The comment seems to be wrong. I'll change it from `the whole codeblock` to `the text event`.

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+
+    while let Some(last) = buf.last() {
+        match last.1 {
+            Event::Text(CowStr::Borrowed(text))
+                if text.trim_matches(&[' ', '\t'] as &[_]).is_empty() =>
+            {
+                let last = buf.pop().unwrap();
+                ret.push(last);
+            }
+            _ => break,
```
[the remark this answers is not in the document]
dalance: Thanks. I'll use `as_ref()`.

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+                                }
+                            };
+
+                            let text = &text[range];
+
+                            // Whitespaces between translate texts should be added to translate
+                            // group.
+                            // So all whitespaces are added to the translate buffer temporary, and
+                            // the trailing whitespaces will be remvoed finally.
+                            let is_whitespace = text.trim_matches(&[' ', '\t'] as &[_]).is_empty();
```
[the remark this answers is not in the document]
dalance: I intend that this is whitespace without linebreak.
For example, I want that the following line comments are grouped into 2 messages.
If `trim()` is used, 5 lines including linebreak only line become a message.

```
// a
// b

// a
// b
// c
```

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+    // Language detection from language identifier of codeblock.
+    let ss = SyntaxSet::load_defaults_newlines();
+    let syntax = if let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(x)))) = &events[0] {
+        ss.find_syntax_by_token(x.split(',').next().unwrap())
+    } else {
+        None
+    };
+
+    let is_translate_scope = |x| {
+        let scope_string = Scope::new("string").unwrap();
```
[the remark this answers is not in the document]
dalance: I'll use `OnceLock` to reuse them.

i18n-helpers/src/lib.rs
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+
+                for (range, scope) in ScopeRangeIterator::new(&ops, text) {
+                    // Calculate line number of the range
+                    let range_line = if range.start == 0 {
+                        *text_line
+                    } else {
+                        text_line + text[0..range.start].lines().count() - 1
+                    };
+
+                    let _ = stack.apply_with_hook(scope, |op, scopes| {
```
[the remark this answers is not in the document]
dalance: I misunderstood the usage of `apply_with_hook`.
If `apply_with_hook` is used, the hook will be called by stack push/pop.
Therefore if the op is `Pop(3)`, the hook is called 3 times. This is not my intention.

I found the mistake through fuzz test, and use `apply` instead of it now.

>  Is it because we're trying to compute line number as well?

Yes

dalance:
I refrected the review, introduced `Flags` struct to resolve the normarilzation issue, and rebased to the latest main.

dalance:
My solution seems to be broken...
`msginit` can't handle flags in messages.pot.
Another solution may be required.

dalance:
Built-in flags seems to be passed through `msginit`.
So using `c-format` flag instead of `codeblock` flag may resolve the issue.
But it may be confusable.

mgeisler:
> I refrected the review, introduced `Flags` struct to resolve the normarilzation issue, and rebased to the latest main.

If this is only for normalization, then please don't add extra code to support it — normalization is something translators (or @henrif75) will run only once after a new breaking release. I don't think the main code should be expanded to support it.

In this case, I believe the problem are some newlines which are incorrect after running normalize? As long as the normal flow works, then these messages will be corrected by `msgmerge` and that's good enough.

@dyoo, are you happy with the logic here? Please finalize the review so we can merge it.

dalance:
If there is no need to resolve the normalization issue, I'll remove the logic.
I think migration from v0.2 to v0.3 will works fine without it.

dalance:
I removed the logic related to `Flags`.

mgeisler:
> If there is no need to resolve the normalization issue, I'll remove the logic.

Thanks @dalance!

> > @henrif75, when this is merged, we will do a 0.3 release. This is a breaking change: we need to run `mdbook-i18n-normalize` on our translations to keep any translated comments intact (as tested by @dalance in [#109 (comment)](https://github.com/google/mdbook-i18n-helpers/pull/109#issuecomment-1784795489)). I'll make sure to explain this again in the release notes.
> 
> Sounds good. I assume it's going to break ongoing PRs?

Hey @henrif75, yes, it will affect PRs that touches code blocks. Most PRs don't, I believe so the impact should be less than when the tooling learnt to ignore most Markdown formatting in version 0.2.

Note that there will only be conflicts if we run the normalization tool on a translation — until we do, the translation will work just fine in all places except for the code blocks.

So I suggest that we land the current in-flight PRs, run the normalization tool, and then people can continue translating like normal (but with much smaller PO files: a quick test shows that we go from 19k to 16k lines!).

i18n-helpers/src/lib.rs
```diff
@@ -334,18 +336,129 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+    // Language detection from language identifier of codeblock.
+    let ss = SyntaxSet::load_defaults_newlines();
+    let syntax = if let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(x)))) = &events[0] {
+        ss.find_syntax_by_token(x.split(',').next().unwrap())
+    } else {
+        None
+    };
+
+    let Some(syntax) = syntax else {
+        // If there is no language specifier, the whole codeblock should be translated.
```
mgeisler, line 372: I just tested this and a quick look at the diff before/after this change shows something interesting:

````patch
+#: src/running-the-course/course-structure.md:57
+msgid ""
+"```shell\n"
+"cargo init concurrency\n"
+"cd concurrency\n"
+"cargo add tokio --features full\n"
+"cargo run\n"
+"```"
+msgstr ""
````

This seemed odd at first: why do we suddenly include a full code block?

The answer is in this line: before we used a simple heuristic and concluded that we could skip this block because it contains neither `"` nor `//`. Now, it is included verbatim because `shell` apparently isn't a recognized syntax name.

@dalance, could you restore the old heuristic in this case?

Comment (dalance) [reply i18n-helpers/src/lib.rs:372]: I'll restore heuristic.

**Comment by dalance** [comment]
I restored heuristic logic for codeblock which doesn't have lang-specifier, and refactored stack operation logic.

File: i18n-helpers/src/lib.rs (hunk 16 above)
Comment (mgeisler) [reply i18n-helpers/src/lib.rs]: Thanks a lot!

File: i18n-helpers/src/bin/mdbook-gettext.rs (hunk 6 above)
Comment (mgeisler) [reply i18n-helpers/src/bin/mdbook-gettext.rs]: No immediate idea here so let's resolve this for now. I would love to land this and then we'll try it out for a bit before releasing it.

File: i18n-helpers/src/lib.rs (hunk 8 above)
Comment (mgeisler) [reply i18n-helpers/src/lib.rs]: Let's do this after merging the PR.

File: i18n-helpers/src/lib.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (mgeisler) [reply i18n-helpers/src/lib.rs]: This looks fixed to me.

File: i18n-helpers/src/lib.rs (hunk 13 above)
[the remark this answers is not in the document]
Comment (mgeisler) [reply i18n-helpers/src/lib.rs]: Splitting the two groups of comments into two messages seems good to me.

File: i18n-helpers/src/lib.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (mgeisler) [reply i18n-helpers/src/lib.rs]: This is now done!

File: i18n-helpers/src/lib.rs [hunk 17]
```diff
@@ -333,18 +335,139 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+
+                let Ok(ops) = ps.parse_line(text, &ss) else {
+                    // If parse is failed, the whole codeblock should be translated.
+                    ret.push(Group::Translate(events[idx..idx + 1].into()));
+                    continue;
+                };
+
+                let mut translate = vec![];
+
+                for (range, scope) in ScopeRangeIterator::new(&ops, text) {
```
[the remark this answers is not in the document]
mgeisler: The variable was renamed to `op`.

i18n-helpers/src/lib.rs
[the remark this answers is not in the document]
mgeisler: I think we can resolve this?

i18n-helpers/src/lib.rs
```diff
@@ -334,18 +336,129 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+
+                for (range, op) in ScopeRangeIterator::new(&ops, text) {
+                    // Calculate line number of the range
+                    let range_line = if range.start == 0 {
+                        *text_line
+                    } else {
+                        text_line + text[0..range.start].lines().count() - 1
+                    };
+
+                    let _ = stack.apply(op);
```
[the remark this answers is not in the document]
mgeisler: This was changed to `stack.apply(op).is_err()`.

i18n-helpers/src/lib.rs
```diff
@@ -334,18 +336,129 @@ fn is_comment_skip_directive(html: &str) -> bool {
...
+
+                let mut translate = vec![];
+
+                for (range, op) in ScopeRangeIterator::new(&ops, text) {
+                    // Calculate line number of the range
+                    let range_line = if range.start == 0 {
+                        *text_line
+                    } else {
+                        text_line + text[0..range.start].lines().count() - 1
+                    };
```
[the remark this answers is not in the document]
mgeisler: It looks like this was moved?

dyoo (approved):
Thanks for the changes; looks good to me.

mgeisler:
Amazing, thanks @dyoo for the review and thanks @dalance for all the work here! I'm very excited to merge and release this to the world :smile:

mgeisler:
Sorry for the tiny merge conflict in `lib.rs`, @dalance :slightly_smiling_face: We should be able to merge this right after you resolve it.

dalance:
I resolved the conflict.

i18n-helpers/fuzz/fuzz_targets/group_events.rs
```diff
@@ -11,8 +11,13 @@ fuzz_target!(|text: String| {
...
         })
-        .cloned()
         .collect::<Vec<_>>();
 
-    assert_eq!(events, flattened_groups);
+    // Comparison through markdown text to detect missing text.
+    // Events can't be compared directly because `group_events`
+    // may split a event into some events.
+    let text_from_events = reconstruct_markdown(&events, None);
+    let text_from_groups = reconstruct_markdown(&flattened_groups, None);
```
mgeisler, line 20: Thanks for this, this is smart!

PR #111, opened 2023-11-05
Skip extracting whitespace-only messages

### LINKED ISSUE #64: Filter out empty `msgid` entries
I tried running `mdbook-xgettext` on the [Rust Book](https://github.com/rust-lang/book). After removing formatting from the SUMMARY.md file, I end up with a `messages.pot` file which is almost correct:

```shell
% msgcat -o po/messages.pot po/messages.pot
po/messages.pot:1660: duplicate message definition...
po/messages.pot:3: ...this is the location of the first definition
msgcat: found 1 fatal error
```

The problem is this entry:

```gettext
#: src/appendix-04-useful-development-tools.md:120
#: src/appendix-04-useful-development-tools.md:149
msgid ""
msgstr ""
```

which in turn originates from a [non-empty HTML tag](https://github.com/rust-lang/book/blob/main/src/appendix-04-useful-development-tools.md?plain=1#L149):

```markdown
<span class="filename">Filename: src/main.rs</span>
```

The empty `msgid` is also produced by [empty HTML tags](https://github.com/rust-lang/book/blob/main/src/ch00-00-introduction.md?plain=1#L169):

```markdown
<span id="ferris"></span>
```

We should handle such entries correctly. This would mean

1. Ensure we don't add duplicate entries to the PO file. This is perhaps something that need fixing in [polib](https://github.com/brettdong/polib).
2. Perhaps we should extract the text inside inline HTML tags such as the `span` above?

mgeisler:
This skips empty messages and messages consisting of whitespace only.

Fixes #64.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -463,7 +463,11 @@ pub fn extract_messages(document: &str) -> Vec<(usize, String)> {
             Group::Translate(events) => {
                 if let Some((lineno, _)) = events.first() {
                     let (text, new_state) = reconstruct_markdown(events, state);
-                    messages.push((*lineno, text));
+                    // Skip empty messages since they are special:
+                    // they contains the PO file metadata.
+                    if !text.trim().is_empty() {
+                        messages.push((*lineno, text));
+                    }
                     state = Some(new_state);
                 }
             }
@@ -687,6 +691,16 @@ mod tests {
     #[test]
     fn extract_messages_empty() {
         assert_extract_messages("", vec![]);
+    }
+
+    #[test]
+    fn extract_messages_empty_html() {
+        assert_extract_messages("<span></span>", vec![]);
+    }
+
+    #[test]
+    fn extract_messages_whitespace_only() {
+        assert_extract_messages("<span>  </span>", vec![]);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


mgeisler:
With this, we can now extract messages from the Rust Book and the second commits implements a test for this!

i18n-helpers/src/lib.rs
```diff
@@ -463,7 +463,10 @@ pub fn extract_messages(document: &str) -> Vec<(usize, String)> {
             Group::Translate(events) => {
                 if let Some((lineno, _)) = events.first() {
                     let (text, new_state) = reconstruct_markdown(events, state);
-                    messages.push((*lineno, text));
+                    if !text.trim().is_empty() {
+                        // Skip empty messages.
```
dyoo, line 467: Maybe reverse the comment's focus to `// Include messages with content.` I briefly got confused because the test was checking for the presence of content, not its absence.

dyoo (approved):

i18n-helpers/src/lib.rs
mgeisler: Thanks, that's a good point! I rephrased it to actually explain _why_ we skip the messages, not just _that_ we skip them.

PR #115, opened 2023-11-08
Test extracting text from multiple books

mgeisler:
This expands our testing to test extracting text from both the Rust Book and Comprehensive Rust. We only check the format of the generated POT file right now, but we might be able to expand this to other checks later.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,12 +25,21 @@ jobs:
       - name: Test
         run: cargo test
 
-  rust-book:
-    name: Test with Rust Book
+  test-other-books:
+    strategy:
+      matrix:
+        include:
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @dyoo, I could use a similar matrix technique to run the fuzz tests in parallel. That would halve the running time of that test.

dyoo (approved):
Looks reasonable to me, but to be honest, I'm not familiar with github workflows!

mgeisler:
> Looks reasonable to me, but to be honest, I'm not familiar with github workflows!

Not many people are :smile: But you see the two jobs on the [checks tab](https://github.com/google/mdbook-i18n-helpers/pull/115/checks) and so you can inspect that it works as intended.

PR #119, opened 2023-11-09
Drop unnecessary hashes from raw strings.

dyoo:
We drop some unnecessary escapes from the raw strings.  This is to clean up some lint warnings from Clippy.


mgeisler (approved):
Thanks!

PR #120, opened 2023-11-09
Apply many pedantic Clippy lints

mgeisler:


mgeisler:
@dyoo, funny enough, we had the same idea here... I played around and enabled a bunch of other lints to clean things up a bit :smile:

dyoo:
Ha;  yeah, I was doing some Clippy runs through Emacs and got annoyed when it was blaming me for things I didn't do.  :P

dyoo (approved):

PR #122, opened 2023-11-11
Create `RenderContext` using a real book

mgeisler:
Instead of hard-coding a large JSON string, we can use the real infrastructure instead and setup an actual book, load it, and use it to generate a context.

Using the actual dependency is preferable to using a mocked dependency since it allows us to test against the real implementation.


mgeisler:
This swaps a `tempdir` dependency for one on `tempfile`: we already use `tempfile` in the rest of the codebase and `tempfile` is still maintained (https://lib.rs/crates/tempdir says that `tempdir` is unmaintained).

sakex (approved):

mgeisler:
Thanks for the quick review, I better take out the `dbg!` statements before merging this! :smile:

PR #123, opened 2023-11-11
Ensure we run tests and Clipply on entire workspace

mgeisler:
Before, we did not actually run tests for `mdbook-tera-backend`. Now that the crate is a default workspace member, we will automatically run tests and Clippy for it too.

This also makes it easier to apply the same lints at home as in CI: just run `cargo clippy` with a nightly Rust compiler. The `lints` table should [become stable with Rust 1.74](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#lints).

### DIFF
--- a/i18n-helpers/fuzz/Cargo.toml
+++ b/i18n-helpers/fuzz/Cargo.toml
@@ -7,6 +7,11 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
+[lints.clippy]
+dbg_macro = "warn"
+print_stdout = "warn"
+print_stderr = "warn"
+
 [dependencies]
 arbitrary = { version = "1.3.1", features = ["derive"] }
 libfuzzer-sys = "0.4.0"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  RUSTFLAGS: "-Dwarnings"
   FORCE_COLOR: 3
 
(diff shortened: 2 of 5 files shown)


kdarkhan:
Using `workspace.lints` table in the root `Cargo.toml` as documented [here](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table) might be a cleaner way to do this.

mgeisler:
> Using `workspace.lints` table in the root `Cargo.toml` as documented [here](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table) might be a cleaner way to do this.

Thanks, this is nice! I've updated it as you suggested.

i18n-helpers/fuzz/Cargo.toml
```diff
@@ -7,6 +7,11 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
+[lints.clippy]
```
kdarkhan, line 10: This does not work because `fuzz` dir is not a part of the workspace but we can make it part of the workspace once a new version of `cargo-fuzz` released which contains https://github.com/rust-fuzz/cargo-fuzz/pull/357

kdarkhan (approved):

i18n-helpers/fuzz/Cargo.toml
mgeisler: Nice! I first considered making the fuzz tests part of the workspace, but it seemed easier to leave it like this. It's cool to hear that we'll soon be able to merge everything.

PR #128, opened 2023-12-08
Define fuzzer for xgettext binary

### LINKED ISSUE #57: Add a fuzz tests for all binaries

kdarkhan:
This defines the last binary fuzzer as a part of https://github.com/google/mdbook-i18n-helpers/issues/57.

Fixes #57.


kdarkhan:
Ping @mgeisler, let me know if you have any comments on this PR.

PR #129, opened 2023-12-21
Fix codeblock in codeblock

dalance:
Codeblock which contains ```` ``` ```` will be like below:

`````markdown
````markdown
```rust
let a = 0;
```
````
`````

But the output of mdbook-i18n-helpers seems to be broken like below:

`````markdown
```markdown
```rust
let a = 0;
```
```
`````

This is because codeblock token count is fixed as 3 in `reconstruct_markdown`.
This PR adds to check the appropriate codeblock token count.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -519,10 +519,11 @@ pub fn reconstruct_markdown(
     group: &[(usize, Event)],
     state: Option<State<'static>>,
 ) -> (String, State<'static>) {
+    let code_block_token_count = check_code_block_token_count(group);
     let events = group.iter().map(|(_, event)| event);
     let mut markdown = String::new();
     let options = Options {
-        code_block_token_count: 3,
+        code_block_token_count,
         list_token: '-',
         emphasis_token: '_',
         strong_token: "**",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mgeisler** [comment]
Thanks @dalance for finding and fixing this! I believe the correct place to fix this would actually be in pulldown-cmark-to-cmark. I went looking there and https://github.com/Byron/pulldown-cmark-to-cmark/issues/20 looks related.

I'm happy to fix this here for now, but we should add a TODO comment with a reference to pulldown-cmark-to-cmark to see if we can fix it for everybody.

File: i18n-helpers/src/lib.rs [hunk 1]
```diff
@@ -556,6 +557,39 @@ pub fn reconstruct_markdown(
     (String::from(markdown.trim_start_matches('\n')), new_state)
 }
 
+/// Check appropriate codeblock token count
+fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
```
mgeisler, line 561:
```suggestion
/// Check appropriate codeblock token count
///
/// See https://github.com/Byron/pulldown-cmark-to-cmark/issues/20
/// for a related upstream issue.
fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
```

mgeisler, line 561:
```suggestion
/// Check appropriate codeblock token count
///
/// See https://github.com/Byron/pulldown-cmark-to-cmark/issues/20
/// for a related upstream issue.
fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
```

mgeisler (approved):

dalance:
@mgeisler Thank you for merging this PR.
Could you release 0.3.1 including this?
(Or is there any release plan?)

This is because the problem is a blocking issue of translation on https://github.com/rust-lang/rust-by-example.

mgeisler:
> @mgeisler Thank you for merging this PR. Could you release 0.3.1 including this? (Or is there any release plan?)
> 
> This is because the problem is a blocking issue of translation on https://github.com/rust-lang/rust-by-example.

Thanks for reminding me of this! Let me create a release now!

mgeisler:
I've made a 0.3.1 release! :tada:

PR #142, opened 2024-01-05
Share common workspace dependencies in root Cargo.toml

kdarkhan:
This approach makes dependency version management between workspace members cleaner as documented in
https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-dependencies-table


mgeisler (approved):
Thanks, that looks neat!

PR #147, opened 2024-01-16
Cache `SyntaxSet` between `parse_codeblock` calls

mgeisler:
Before we would load and deserialize the `syntect` syntax trees (Sublime Text syntax definitions) on every call. We now cache this read-only data.

Before, `mdbook build` took 1.44 s, now it takes 0.56 s instead. I started looking in to optimizing this because I'm considering making `mdbook-xgettext` an (optional) plugin in `book.toml` for Comprehensive Rust. This will imply that we run the plugin on every `mdbook build` — at least for people like me who have the plugin installed.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -396,7 +396,9 @@ fn heuristic_codeblock<'a>(events: &'a [(usize, Event)]) -> Vec<Group<'a>> {
 /// Creates groups by parsing codeblock.
 fn parse_codeblock<'a>(events: &'a [(usize, Event)]) -> Vec<Group<'a>> {
     // Language detection from language identifier of codeblock.
-    let ss = SyntaxSet::load_defaults_newlines();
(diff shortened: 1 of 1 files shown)


kdarkhan (approved):

kdarkhan:
Just a thought - maybe we should add several `bench` targets so that we can find how performance of the code changes. Does not have to be exposed during PR build but something that is easy to trigger locally.

mgeisler:
> Just a thought - maybe we should add several `bench` targets so that we can find how performance of the code changes. Does not have to be exposed during PR build but something that is easy to trigger locally.

That would indeed be a useful tool to have!

For this, I used https://github.com/sharkdp/hyperfine like this:

```
% MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages-10.pot", "granularity": 10}}' hyperfine 'mdbook build -d po'
```

to test only `mdbook-xgettext`. But if we could put this into a `criterion` benchmark as well, then that would be nice.

Issue #150, opened 2024-01-16
`group_events` fuzz test is flaky


mgeisler:
After #129, the `group_events` fuzz test seems to have become flaky. @dalance, would you be able to take a look?

An example failure is [here](https://github.com/google/mdbook-i18n-helpers/actions/runs/7542340221/job/20530931734), where the failing input can be minimized to

```
"````d\n```"
```

The diff of the failure is

```
<    "````d\n```\n````",
>    "```d\n```\n```",
```

meaning that `reconstruct_markdown(&events, None)` returned

~~~markdown
````d
```
````
~~~

whereas `reconstruct_markdown(&flattened_groups, None)` returned

~~~markdown
```d
```
```
~~~

My guess is that this is because the counting of consequetive `` ` `` is slightly off.

Indeed, looking at `flattened_groups` in the fuzz test, I see

```rust
[fuzz_targets/group_events.rs:22:5] &flattened_groups = [
    (
        1,
        Start(
            CodeBlock(
                Fenced(
                    Borrowed(
                        "d",
                    ),
                ),
            ),
        ),
    ),
    (
        2,
        Text(
            Borrowed(
                "`",
            ),
        ),
    ),
    (
        2,
        Text(
            Borrowed(
                "`",
            ),
        ),
    ),
```

with lots of lone `` ` `` characters.

A related question, is this not something which should be fixed in [pulldown-cmark-to-cmark](https://github.com/Byron/pulldown-cmark-to-cmark) instead of here? @dalance, could you create an issue in that repository and see if you can move the fix from here to there?

dalance:
I've opened a PR to fix the existing token couting code.

I agree pulldown-cmark-to-cmark should construct an appropriate markdown without workaround.
I'll consider about it.

mgeisler:
I let it run for a few million iterations on my desktop and it looks rock solid now!

Fixed by #153.

PR #153, opened 2024-01-24
Fix code block token count

dalance:
This PR fixes wrong token count in code block reported at #150.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -569,23 +569,32 @@ fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
     let events = group.iter().map(|(_, event)| event);
     let mut in_codeblock = false;
     let mut max_token_count = 0;
+
+    // token_count should be taken over Text events
+    // because a continuous text may be splitted to several Text events.
+    let mut token_count = 0;
+
     for event in events {
         match event {
-            Event::Start(Tag::CodeBlock(_)) => in_codeblock = true,
-            Event::End(Tag::CodeBlock(_)) => in_codeblock = false,
+            Event::Start(Tag::CodeBlock(_)) => {
+                in_codeblock = true;
+                token_count = 0;
+            }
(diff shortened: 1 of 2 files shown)


i18n-helpers/src/lib.rs
```diff
@@ -569,12 +569,22 @@ fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
     let events = group.iter().map(|(_, event)| event);
     let mut in_codeblock = false;
     let mut max_token_count = 0;
+
+    // token_count should be taken over Text events
+    // because a continuous text may be splitted to some Text events.
```
kdarkhan, line 574: Nit: If this is about splitting a single text into several `Text` events, then `several` instead of `some` might be more accurate in this context.

kdarkhan (approved):

i18n-helpers/src/lib.rs
```diff
@@ -569,12 +569,22 @@ fn check_code_block_token_count(group: &[(usize, Event)]) -> usize {
...
-            Event::End(Tag::CodeBlock(_)) => in_codeblock = false,
+            Event::Start(Tag::CodeBlock(_)) => {
+                in_codeblock = true;
+                token_count = 0;
+            }
+            Event::End(Tag::CodeBlock(_)) => {
+                in_codeblock = false;
+                token_count = 0;
+            }
             Event::Text(x) if in_codeblock => {
```
kdarkhan, line 587: I think this `Event::Text` arm will be easier to understand if `max_token_count` is updated only in a single place:

```
            Event::Text(x) if in_codeblock => {
                for c in x.chars() {
                    if c == '`' {
                        token_count += 1;
                        max_token_count = std::cmp::max(max_token_count, token_count);
                    } else {
                        token_count = 0;
                    }
                }
            }
```

dalance:
@kdarkhan Thank you for your review. I've fixed the comment and `max_token_count` logic.

kdarkhan:
Thanks, looks good!

PR #156, opened 2024-01-27
Check for duplicate sources in messages

### LINKED ISSUE #154: Deduplicate repeated sources
With #145 merged, we can now end up in a situation where a message has the same source repeated multiple times:

```gettext
#: src/SUMMARY.md:10 src/SUMMARY.md:10 src/SUMMARY.md:70 src/SUMMARY.md:90
msgid "Welcome"
msgstr "Te damos la bienvenida"
```

We should add logic to remove the duplicates in this case. The situation becomes even worse when the line numbers are removed completely (when granularity is set to `0`).

zachcmadsen:
With rounded line numbers, it's possible for messages to have duplicate sources. This checks for duplicate sources before updating messages. An alternative approach is deduplicating sources after all the messages are added/updated. I went with the "on-the-fly" approach since it's simpler.

The check scans all of the sources for a message. If performance is a concern, we could try other ways to check for duplicates.

Fixes #154

### DIFF
--- a/i18n-helpers/src/xgettext.rs
+++ b/i18n-helpers/src/xgettext.rs
@@ -404,4 +404,39 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_create_catalog_lineno_granularity_duplicates() -> anyhow::Result<()> {
+        let (ctx, _tmp) = create_render_context(&[
+            (
+                "book.toml",
+                "[book]\n\
+                 [output.xgettext]\n\
+                 granularity = 3",
+            ),
+            ("src/SUMMARY.md", "- [Foo](foo.md)"),
+            (
+                "src/foo.md",
+                "Bar\n\
+                 \n\
+                 Bar\n\
+                 \n\
+                 Bar\n",
+            ),
+        ])?;
+
+        let catalog = create_catalog(&ctx, std::fs::read_to_string)?;
+        assert_eq!(
+            catalog
+                .messages()
+                .map(|msg| (msg.source(), msg.msgid()))
+                .collect::<Vec<_>>(),
+            &[
+                ("src/SUMMARY.md:1", "Foo"),
+                ("src/foo.md:1 src/foo.md:3", "Bar"),
+            ]
+        );
+
+        Ok(())
+    }
 }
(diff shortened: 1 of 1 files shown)


i18n-helpers/src/xgettext.rs
```diff
@@ -41,7 +41,13 @@ fn strip_link(text: &str) -> String {
 
 fn add_message(catalog: &mut Catalog, msgid: &str, source: &str) {
     let sources = match catalog.find_message(None, msgid, None) {
-        Some(msg) => wrap_sources(&format!("{}\n{}", msg.source(), source)),
+        Some(msg) => {
+            if msg.source().contains(source) {
```
kdarkhan, line 45: I think checking for existence should be done per line rather than regular string contains. An example case where the proposed logic would fail will be if source contains entry `some-dir/README.md` and the new source is `README.md`.

mgeisler: Good catch @kdarkhan! I think this could be handled nicely by

1. Separating the sources with `\n` when `add_message` is called.
2. Use `lines()` to split the source field back to lines and then deduplicate these lines.
3. Do a call to `wrap_sources` to nicely wrap things.

Both 2 and 3 could be done _after_ adding all messages to the catalog — I think that would give us linear complexity here. Right now, we have some potential for O(n²) complexity here.

Putting a filename per line remove ambiguity about filenames that contain spaces. I'm not actually sure how such filenames ought to be stored in the sources field — feel free to experiment a bit with `xgettext` if you are curious!

zachcmadsen: Thanks for the outline @mgeisler! I pushed an update along those lines. What I have assumes that duplicate sources are consecutive. Are there cases where that doesn't hold?

mgeisler: > What I have assumes that duplicate sources are consecutive. Are there cases where that doesn't hold?

I think it should hold right now — we walk over the contents of the book in the order given by [`Book::iter`](https://docs.rs/mdbook/latest/mdbook/book/struct.Book.html#method.iter). It says that it's doing a depth-first traversal on the files of the book, so it ought to do the right thing!

kdarkhan: This change looks good to me.

Here is the diff of the generated `messages.pot` from `comprehensive-rust` repo.

https://gist.github.com/kdarkhan/9e270ae1e5842abd10f826dcaa5940aa

kdarkhan (approved):

kdarkhan:
Thanks for contribution @zachcmadsen! Feel free to pick up the followup task https://github.com/google/mdbook-i18n-helpers/issues/171 if you want.

Issue #148, opened 2024-01-16
Code coverage uploads sometimes fail


mgeisler:
Hi @kdarkhan, I've seen a few Codecov uploads fail. An example is in [this run](https://github.com/google/mdbook-i18n-helpers/actions/runs/7529358743). The error is

> [2024-01-15T13:13:27.254Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 503 - upstream connect error or disconnect/reset before headers. reset reason: connection failure
> Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255

Do you have any idea what could trigger this?

kdarkhan:
Right, I saw this yesterday. I could not find a definitive answer but it might be resolved by using a project specific token instead of tokenless authentication. To implement it, somebody in `google` Github org needs to copy the token from [here](https://app.codecov.io/gh/google/mdbook-i18n-helpers/settings) into repository secrets [here](https://github.com/google/mdbook-i18n-helpers/settings/secrets/actions). The name of the secret should be `CODECOV_TOKEN`. Once that is done, I can adjust the workflow file to test it.

mgeisler:
> The name of the secret should be `CODECOV_TOKEN`. Once that is done, I can adjust the workflow file to test it.

Thanks, I added the repository secret.

kdarkhan:
I merged the PR for using the new token. I am not sure if it will resolve this issue. Let's monitor for several weeks and can resolve this if the pipeline is stable.

mgeisler:
Sounds good, thanks for looking into this so quickly!

kdarkhan:
There was [another failure](https://github.com/google/mdbook-i18n-helpers/actions/runs/7617792277/job/20747595520?pr=152). The error message is

> [2024-01-22T21:51:51.270Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

According to [this article](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954), adding the token should improve the situation but will not guarantee 100% success rate.

mgeisler:
[...]

PR #162, opened 2024-01-30
Make codecov report use secret

kdarkhan:
This addresses #148. Codecov coverage report upload are not stable. The suggested solution is to use
per project secret. This project have that enabled but it is not usable for pull request workflows triggered from forks. Due to Github's security restrictions, only PRs based on local branches have access to secrets.

This PR amends the existing workflow by not pushing the reports to Codecov. The reports are instead uploaded to Github artifacts (the artifacts are stored for 90 days by default).

Then a new workflow is introduced which executes after every successful execution of the existing workflow. It downloads the artifact and uploads the report to Codecov using the security token.

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -1,0 +1,60 @@
+name: Coverage report
+
+on:
+  # This workflow is triggered after every successfull execution
+  # of `tests` workflow.
+  workflow_run:
+    workflows: ["tests"]
+    types:
+      - completed
+jobs:
+  coverage:
+    name: Coverage report
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Wow, thanks for implementing this — it looks good and I'm excited to see it in action.

.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,56 @@
+name: Coverage report
+
+on:
+  workflow_run:
+    workflows: ["tests"]
```
mgeisler, line 5: Could you add a comment explaining that this makes the workflow here trigger after every completed run of the `tests` workflow?

kdarkhan: Done.

PR #163, opened 2024-01-30
Define grouping for dependabot updates

kdarkhan:
This is based on the example from official documentation at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-2


mgeisler (approved):
Thanks!

PR #166, opened 2024-01-31
Make dependabot grouping work on all update types

kdarkhan:
The initial version defined groups only for cargo
deps. Because of that, we have 2 separate PRs related to `github-actions`. This ensures that grouping logic works for all deps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,12 +7,18 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    groups:
+      minor:
+        update-types:
+          - minor
+      patch:
+        update-types:
+          - patch
 
   - package-ecosystem: cargo
     directory: /
     schedule:
       interval: weekly
-
     groups:
       minor:
         update-types:


.github/dependabot.yml
```diff
@@ -13,10 +13,10 @@ updates:
...
       interval: weekly
 
-    groups:
-      minor:
-        update-types:
-          - minor
-      patch:
-        update-types:
-          - patch
+groups:
```
mgeisler, line 16: I'm not sure you can use `groups` as a top-level item like this. My impression was that it has to be nested under an item in `updates`. See how `groups` is mentioned under `updates` here: [Configuration options for the dependabot.yml file](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-the-dependabotyml-file).

In Comprehensive Rust, I simply duplicated the `groups` for each type of update.

mgeisler (approved):
Let me approve this now — then you can fix the comment about `groups` if you also agree with my reasoning.

.github/dependabot.yml
kdarkhan: I see. I based my approach on [the example 1](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-1) in the official doc s but it might be just not a full example. I just amended the approach and duplicated the definitions.

PR #168, opened 2024-02-02
Fix flaky codecov reports

kdarkhan:
Currently implementation of codecov reports sometimes attaches the analysis to main branch instead of the triggering PR. The root cause of the issue is a bug in Github which does not always populate `github.event.workflow_run.pull_requests[0]`. When it is populated, the current implementation works. When it is missing, `override_commit` and `override_pr` params to `codecov-action` become unset.

More information available at
https://github.com/orgs/community/discussions/25220

The workaround is to store both commit SHA and PR number in the original workflow since those are available there.

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -12,51 +12,98 @@ jobs:
     name: Coverage report
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.workflow_run.pull_requests[0].head.sha }}
 
       - name: 'Download existing coverage report'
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Sure, I can see how this can be made to work.

Another option (a much simpler option) would be to only trigger the workflow on the `main` branch after a merge.

The prr-PR reports are not super important for me — they can be a little silly and make people focus on a 0.3% drop in coverage when they should really be focusing on writing maintainable code and tests 🙂

kdarkhan:
Yeah, the other option is much simpler. I think having an ability to do this is good though. If we decide to use PR report we now have it. If we don't need it, filtering can be done using `if: ${{ ... }`

Issue #173, opened 2024-02-06
codecov-action upgrade broke coverage reports


kdarkhan:
Coverage reports started failing after [upgrade from v3 to v4](https://github.com/google/mdbook-i18n-helpers/pull/169).

Example failure [here](https://github.com/google/mdbook-i18n-helpers/actions/runs/7804318002/job/21286005311)

Error message

```
==> Running version v0.4.6
==> Running command '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov create-commit'
/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov create-commit -C 4a213d1e5bb64fed6de71036439f53d1c796564e --pr 155 -Z
info - 2024-02-06 18:22:00,803 -- ci service found: github-actions
warning - 2024-02-06 18:22:00,806 -- No config file could be found. Ignoring config.
info - 2024-02-06 18:22:00,951 -- Process Commit creating complete
error - 2024-02-06 18:22:00,951 -- Commit creating failed: ["Service not found: none"]
Traceback (most recent call last):
  File "codecov_cli/main.py", line 81, in <module>
  File "codecov_cli/main.py", line 77, in run
  File "click/core.py", line 1157, in __call__
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1688, in invoke
  File "click/core.py", line 1434, in invoke
  File "click/core.py", line 783, in invoke
  File "click/decorators.py", line 33, in new_func
  File "codecov_cli/commands/commit.py", line 64, in create_commit
  File "codecov_cli/services/commit/__init__.py", line 39, in create_commit_logic
  File "codecov_cli/helpers/request.py", line 133, in log_warnings_and_errors_if_any
NameError: name 'exit' is not defined
[1739] Failed to execute script 'main' due to unhandled exception!
Error: Codecov: Failed to properly create commit: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

mgeisler:
Hey @kdarkhan, I would suggest moving to the simplest setup: only generate covera reports after merges. That should work fine: it will allow us to use the privileged trigger from the `main` branch.

We will still get the history of coverage over time on `main` and we can then look at this to determine where we have big holes in our test coverage.

kdarkhan:
@mgeisler, that can be done and it will simplify commit/PR detection logic I defined in the pipelines.

[...]

PR #179, opened 2024-02-29
Upgrade codecov-action to v4

kdarkhan:
With the issues https://github.com/codecov/feedback/issues/263 and https://github.com/codecov/codecov-action/issues/1287 resolved, we can upgrade to v4 again.

This fixes https://github.com/google/mdbook-i18n-helpers/issues/173

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -95,7 +95,7 @@ jobs:
           path: repo_root
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ${{ github.workspace }}/codecov_report.json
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage-report.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           path: repo_root
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
```
qwandor, line 98: Should this maybe be `v4.1`, if `v4.0` has the bug?

kdarkhan: `codecov-action` maintainers seem to move `v4` tag as they release new minor versions. `v4` points to `v4.1.0` according to [this](https://github.com/codecov/codecov-action/tree/v4). I think it will be simpler to have it as v4 so we don't need to change when v4.2 is released.

mgeisler: > Should this maybe be `v4.1`, if `v4.0` has the bug?

Good question!

> `codecov-action` maintainers seem to move `v4` tag as they release new minor versions. `v4` points to `v4.1.0` according to [this](https://github.com/codecov/codecov-action/tree/v4).

Yes, this is the way tags for GitHub actions typically work: people move them around so that you get the latest semver-compatible version when you depend on a given tag. It's a bit silly, but that's the way things work last time I checked.

The GitHub documentation still recommends this: [Using tags for release management](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management).

mgeisler (approved):
Thanks for tracking this!

Issue #125, opened 2023-11-12
Wrong normalization of message with HTML


mgeisler:
This is from https://github.com/google/comprehensive-rust/issues/1471: running `mdbook-i18n-normalize` on a translated message that contain HTML gives the wrong output.

I can reproduce this with

```patch
diff --git a/i18n-helpers/src/normalize.rs b/i18n-helpers/src/normalize.rs
index 7fed94d..4c0fd5c 100644
--- a/i18n-helpers/src/normalize.rs
+++ b/i18n-helpers/src/normalize.rs
@@ -451,6 +451,15 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_normalize_html() {
+        let catalog = create_catalog(&[("foo <span>bar</span>", "FOO <span>BAR</span>")]);
+        assert_normalized_messages_eq(
+            catalog,
+            &[exact("foo <span>bar</span>", "FOO <span>BAR</span>")],
+        );
+    }
+
     #[test]
     fn test_normalize_disappearing_html() {
         // Normalizing "<b>" results in no messages.
```

This tests normalizing a catalog looking like this:

```po
msgid "foo <span>bar</span>"
msgstr "FOO <span>BAR</span>"
```

The test fails with

```
---- normalize::tests::test_normalize_html stdout ----
thread 'normalize::tests::test_normalize_html' panicked at i18n-helpers/src/normalize.rs:457:9:
assertion failed: `(left == right)`

Diff < left / right > :
 [
     (
         false,
<        "foo ",
<        "FOO ",
<    ),
<    (
<        false,
<        "bar",
<        "BAR",
>        "foo <span>bar</span>",
>        "FOO <span>BAR</span>",
     ),
 ]
```

which tells us that the normalized catalog contain _two_ messages:

```po
msgid "foo"
msgstr "FOO"

msgid "bar"
msgstr "BAR"
```

In other words, the original message was split into two messages.

In case only the translation contain HTML, the normalization will end up with a different number of messages: 1 for the `msgid` field, and 2 for the `msgstr` field. This is seen as an error, so a fallback kicks in: the normalized message is marked fuzzy and we accumulate the "left-over" messages into the final message.

This behavior can be seen in this unit test in `normalize.rs`:

```rust
    #[test]
    fn test_normalize_fuzzy_list_items_too_many() {
        let catalog = create_catalog(&[(
            "* foo\n\
             * bar",
            "* FOO\n\
             * BAR\n\
             * BAZ",
[...]

PR #183, opened 2024-04-11
Upgrade pulldown-cmark to 0.10.3

kdarkhan:
The latest version of `pulldown-cmark-to-cmark` now supports `pulldown-cmark`. This should allow us to upgrade both dependencies.

Had to adjust TagEnd handling and change some types to fix the build failures.

Information on breaking changes in `pulldown-cmark` is available [here](https://github.com/pulldown-cmark/pulldown-cmark/releases). Not sure if I am missing anything. I hope that the tests we have are extensive enough to catch any issues.

### DIFF
--- a/i18n-helpers/src/gettext.rs
+++ b/i18n-helpers/src/gettext.rs
@@ -310,7 +310,7 @@ mod tests {
         let catalog = create_catalog(&[("Foo", "FOO"), ("Bar", "BAR")]);
         assert_eq!(
             translate("# Foo { #id .foo }", &catalog),
-            "# FOO {#id .foo}"
+            "# FOO { #id .foo }"
         );
     }
 
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -24,7 +24,9 @@
 //! how to use the supplied `mdbook` plugins.
 
 use polib::catalog::Catalog;
-use pulldown_cmark::{CodeBlockKind, Event, LinkType, Tag};
+use pulldown_cmark::{
+    BrokenLinkCallback, CodeBlockKind, DefaultBrokenLinkCallback, Event, LinkType, Tag, TagEnd,
+};
 use pulldown_cmark_to_cmark::{
     calculate_code_block_token_count, cmark_resume_with_options, Options, State,
 };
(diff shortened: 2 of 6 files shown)


i18n-helpers/src/gettext.rs
```diff
@@ -278,7 +278,7 @@ mod tests {
         ]);
         assert_eq!(
             translate("A footnote[^note].\n\n[^note]: More details.", &catalog),
-            "A FOOTNOTE[^note].\n\n[^note]: MORE DETAILS."
+            "A FOOTNOTE\\[^note\\].\n\n[^note]: MORE DETAILS."
```
mgeisler, line 281: I think this looks wrong since it looks like the translated text disables the foot note?

i18n-helpers/src/lib.rs
```diff
@@ -311,7 +334,7 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
                 }
             }
 
-            Event::Html(s) => {
+            Event::Html(s) | Event::InlineHtml(s) => {
```
mgeisler, line 337: Nice, this is the new kind of HTML you mentioned to me?

I think this is the right thing to do right now, but I wonder if we can make good use of this to deal with #125?

mgeisler (commented):
Thanks a lot for tackling this! I thin the footnote handling is off now, but the rest looks great!

i18n-helpers/src/lib.rs
kdarkhan: Yes, that what I was thinking about as well. Can be done once this PR is merged.

i18n-helpers/src/gettext.rs
kdarkhan: Yeah, it does look wrong. I will try to tackle it this week.

i18n-helpers/src/lib.rs
mgeisler: That sounds great!

i18n-helpers/src/gettext.rs
kdarkhan: Sorry for taking so long. Was able to finally get back to this.

The new version of pulldown-cmark introduced a new way to parse footnotes which is incompatible with the old way. It is still possible to enable the old approach by changing options which I did. The tests are resolved now.

I am not exactly sure why the new approach breaks but I believe it is because the footnote is not parsed as footnote in case it is not able to find the footnote definition. When we split up the events into groups the footnote definition becomes part of a separate group and because of that it seems to break.

mgeisler: Alright, thanks for digging into this! I guess either way is fine for us — long-term, we should probably adapt to the new way, in case the old way gets removed. We can deal with that if or when it happens.

i18n-helpers/src/normalize.rs
```diff
@@ -49,11 +49,14 @@ fn has_broken_link(text: &str) -> bool {
...
-                _,
-                _
-            ))
+            Event::Start(Tag::Link {
+                link_type: LinkType::ReferenceUnknown
+                    | LinkType::CollapsedUnknown
+                    | LinkType::ShortcutUnknown,
+                dest_url: _,
+                title: _,
+                id: _
```
mgeisler, line 58: I think this could be replaced with `..` like you do below. It's of course not an important difference :smile:

mgeisler (approved):
Thanks a lot @kdarkhan!

i18n-helpers/src/normalize.rs
kdarkhan: Agreed, done.

PR #195, opened 2024-05-17
Keep Inline HTML tags in the translated text group while ignoring block level HTML tags

### LINKED ISSUE #125: Wrong normalization of message with HTML

michael-kerscher:
Fixes #125 by keeping inline HTML tags in the translation text group but skip block level HTML tags.
This is possible with the upgrade of pulldown-cmark in #183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,6 +113,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install nightly Rust
+        # TODO: Unpin this after 2024-06-10
         run: rustup default nightly-2024-05-10
 
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,8 +113,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install nightly Rust
-        # TODO: Unpin this after 2024-06-10
-        run: rustup default nightly-2024-05-10
+        run: rustup default nightly
 
       - uses: Swatinem/rust-cache@v2
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,7 +113,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install nightly Rust
-        run: rustup default nightly
+        run: rustup default nightly-2024-05-10
 
       - uses: Swatinem/rust-cache@v2
         with:
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -368,25 +368,31 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
                         ctx.comments.push(comment);
                     }
                     _ => {
-                        // Otherwise, treat as a skipping group if this is a block level Html tag
-                        if matches!(event, Event::Html(_)) {
-                            if let State::Translate(_) = state {
-                                let mut next_groups;
-                                (next_groups, ctx) = state.into_groups(idx, events, ctx);
-                                groups.append(&mut next_groups);
+                        match event {
+                            Event::Html(_) => {
+                                // Otherwise, treat as a skipping group if this is a block level Html tag
+                                if let State::Translate(_) = state {
+                                    let mut next_groups;
+                                    (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                                    groups.append(&mut next_groups);
 
-                                state = State::Skip(idx);
+                                    state = State::Skip(idx);
+                                }
                             }
-                        } else if matches!(event, Event::InlineHtml(_)) {
+                            Event::InlineHtml(_) =>
                             // If we're currently skipping, then a new
                             // translatable group starts here.
-                            if let State::Skip(_) = state {
-                                let mut next_groups;
-                                (next_groups, ctx) = state.into_groups(idx, events, ctx);
-                                groups.append(&mut next_groups);
+                            {
+                                if let State::Skip(_) = state {
+                                    let mut next_groups;
+                                    (next_groups, ctx) = state.into_groups(idx, events, ctx);
+                                    groups.append(&mut next_groups);
 
-                                state = State::Translate(idx);
+                                    state = State::Translate(idx);
+                                }
                             }
+                            // this code is inside a match of Event::{Html|InlineHtml}, other types are not possible
+                            _ => unreachable!(),
                         }
                     }
                 }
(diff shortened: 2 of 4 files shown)


i18n-helpers/src/lib.rs
```diff
@@ -367,14 +367,26 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Vec<Group<'a>> {
...
-                    // Otherwise, treat as a skipping group.
                     _ => {
-                        if let State::Translate(_) = state {
-                            let mut next_groups;
-                            (next_groups, ctx) = state.into_groups(idx, events, ctx);
-                            groups.append(&mut next_groups);
-
-                            state = State::Skip(idx);
+                        // Otherwise, treat as a skipping group if this is a block level Html tag
+                        if matches!(event, Event::Html(_)) {
```
mgeisler, line 372: Could the `matches!` macro here be moved up into the `match` statement above?

That is, instead of a single match arm with

```rust
Event::Html(s) | Event::InlineHtml(s) => {
```

above, have separate `Event::Html(s)` and `Event::InlineHtml(s)` arms.

In general, the `matches!` macro is used rather rarely in Rust since the `match` or `if let` statements normally allow you to do the same thing.

michael-kerscher: I was hesitant to duplicate the directives::find() call and handling of Skip and Comment.
I'll give it another pass

kdarkhan:
The fuzz step is failing due to some changes in rust nightly. The failure seems to occur from one of the dependencies.

Until the dependency is fixed, you can temporarily pin to specific nightly version by changing [this line](https://github.com/google/mdbook-i18n-helpers/blob/5b855b0eaf73f91b6a430a41baadf0ae846740f0/.github/workflows/test.yml#L116) to `rustup default nightly-2024-05-10`.

kdarkhan:
The rules that determine whether the HTML is block or inline surprised me. I naively assumed that if a piece of HTML is on a separate line, then it is a block, and inline HTML otherwise. Turns out I was wrong. The rules are defined in https://spec.commonmark.org/0.31.2/#html-blocks. Some tags like `p` or `div` in a single line are considered to be blocks, while `span` is an inline.

Here is an [example](https://spec.commonmark.org/dingus/?text=%3Cdiv%3EThis%20is%20block%20because%20it%20is%20a%20div%3C%2Fdiv%3E%0A%0A%3Cspan%3EThis%20is%20inline%3C%2Fspan%3E%0A%0A%3Cspan%3E%3Cdiv%3EThis%20div%20is%20inline%3C%2Fdiv%3E%3C%2Fspan%3E%0A%0A%3Cspan%3E%0AThis%20is%20block%0A%3C%2Fspan%3E%0A%0AThis%20div%20%3Cdiv%3Enot%20block%3C%2Fdiv%3E%20because%20it%20is%20not%20starting%20the%20line.) which demonstrate how they differ. Switch to `AST` tab to view inlines vs blocks.

I believe even with above rules, not splitting inline elements into separate groups is an improvement.

i18n-helpers/src/lib.rs
mgeisler: Ah, I see! I didn't think of that :smile: 

I think changing the two `matches!` calls here to a nested `match` statement would be a small improvement. The whole logic here is becoming a little complex and it could very well be ripe for some refactoring (in a later PR).

kdarkhan (approved):

.github/workflows/test.yml
```diff
@@ -113,7 +113,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install nightly Rust
-        run: rustup default nightly
+        run: rustup default nightly-2024-05-10
```
mgeisler, line 116: I think it would be good to remind ourselves to remove this. Otherwise others will likely think that this is there for a specific reason.

Let's hope it's fixed in a month :smile: 

```suggestion
        # TODO: Unpin this after 2024-06-10
        run: rustup default nightly-2024-05-10
```

mgeisler (approved):
Thanks @michael-kerscher!

.github/workflows/test.yml
kdarkhan: I just merged an update to `time` dep in https://github.com/google/mdbook-i18n-helpers/pull/198 and the workaround is no longer necessary.

michael-kerscher: I switched back to "nightly" only

PR #198, opened 2024-05-21
Execute `cargo update`

kdarkhan:
This fixes the compilation issue in `time` crate
by upgrading it from 0.3.30 to 0.3.36.

Failure example https://github.com/google/mdbook-i18n-helpers/actions/runs/9171277226/job/25215262973.

If less diff is desired, I could update only `time` by `cargo update time` but probably we can update all minor versions.


mgeisler (approved):
Thanks a lot, a full update is great!

PR #205, opened 2024-06-24
Upgrade pulldown-cmark and pulldown-cmark-to-cmark

kdarkhan:
* [pulldown-cmark/pulldown-cmark@v0.10.3...v0.11.0](https://github.com/pulldown-cmark/pulldown-cmark/compare/v0.10.3...v0.11.0)
* [Byron/pulldown-cmark-to-cmark@v13.0.0...v15.0.1](https://github.com/Byron/pulldown-cmark-to-cmark/compare/v13.0.0...v15.0.1)


rustfmt.toml
```diff
@@ -1,2 +0,0 @@
-# https://github.com/rust-lang/rustfmt/issues/4991
-imports_granularity = "Module"
```
kdarkhan, line 2: Nightly `cargofmt` is failing dprint check due to printing to `stderr` strings like `rewriting static`. It does not seem to introduce any diff in formatting but dprint github action seems to get confused by those logs.

I adjusted the action to use stable `cargofmt` and it seems to produce no diff either. So maybe we can switch to stable version? @mgeisler

mgeisler: Yeah, this is fine for now! The reason we don't see any diff is that the default is `Preserve`.

mgeisler (approved):

PR #209, opened 2024-07-04
i18n-report: utility to generate HTML reports of translation stats

qwandor:


mgeisler:
Hey @qwandor, this is really cool!

One idea: this binary could be it's own crate since I don't think it's dependant on mdbook like the rest of the binaries in mdbook-i18n-helpers?

That way we won't get a dependency on Tera for everybody who uses the translation infrastructure here. The crate could live as a sibling in the same workspace.

qwandor:
> One idea: this binary could be it's own crate since I don't think it's dependant on mdbook like the rest of the binaries in mdbook-i18n-helpers?

Done.

mgeisler:
Awesome, thanks @qwandor! Would you do the honors and publish it to crates.io?

qwandor:
Yep, I'm off today but I'll do that on Monday.

mgeisler:
Btw, I squash-merged it since there branch had commits which fix warnings and do refactors. I hope that's okay with you! I prefer to do normal merges since it's better for both contributors and the history, but that requires contributors to force-push a clean history.

PR #210, opened 2024-07-05
Add i18n-report mode to show diff between two sets of translations

qwandor:
The report looks like this:

```
Counts are "translated (fuzzy, fuzzy untranslated) / total"
Removed bn: 752 (54, 1) / 3761, creation date 2024-03-03T20:37:57+05:30
Added fo: 568 (178, 1) / 3651, creation date 2024-01-24T13:24:49+01:00
Changed ro: 1032 (2731, 0) / 3763, creation date 2024-05-15T20:58:33+03:00 -> ro: 1018 (2731, 0) / 3749, creation date 2024-05-15T20:58:33+03:00
```

The intention is that this can be used from a GitHub action to show a summary of translation changes on PRs.

### DIFF
--- a/i18n-report/src/main.rs
+++ b/i18n-report/src/main.rs
@@ -17,24 +17,50 @@
 
 mod stats;
 
-use anyhow::{bail, Context as _};
+use anyhow::Context as _;
+use clap::Parser;
 use polib::po_file;
 use stats::MessageStats;
-use std::{fs, path::Path};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 use tera::{Context, Tera};
 
 const REPORT_TEMPLATE: &str = include_str!("../templates/report.html");
 
 fn main() -> anyhow::Result<()> {
-    let args = std::env::args().collect::<Vec<_>>();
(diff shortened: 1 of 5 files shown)


i18n-report/src/main.rs
```diff
@@ -54,3 +85,89 @@ fn main() -> anyhow::Result<()> {
...
+    let mut new_translations = all_stats(&po_files(new_translations_directory)?)?;
+    old_translations.sort_by_key(|stats| stats.language.clone());
+    new_translations.sort_by_key(|stats| stats.language.clone());
+
+    if old_translations == new_translations {
+        return Ok(());
+    }
+
+    let mut old_iter = old_translations.iter();
+    let mut new_iter = new_translations.iter();
```
kdarkhan, line 118: I am not very familiar with the workflows in the main repo and the way this tool is used there. Is it common that new languages are being added and removed? If that happens, I assume this report might missmatch PO files for different languages. Is that expected?

kdarkhan: ah, never mind, I just read the implementation and understood how it is done.

kdarkhan (approved):

PR #212, opened 2024-07-09
Advance POT-Creation-Date by 1 minute on normalization

### LINKED ISSUE #117: Advance POT-Creation-Date by 1 second in `mdbook-i18n-normalize`
When `mdbook-i18n-normalize` is used on a PO file, we currently leave the metadata alone. However, this can cause problems for publishing pipelines like the one in Comprehensive Rust: there we expect that the POT-Creation-Date field has changed if any of the `msgid` fields change.

To handle this situation in a smooth way, I propose incrementing the timestamp by 1 second, if possible. This means: parse the field, increment it by 1 second if it's a valid timestamp, and finally write it back to the file. If it's missing or invalid, then leave it along.

carreter:
Fixes #117.

Advances POT-Creation-Date by 1 minute rather than 1 second as @mgeisler suggested as it seems like the date format does not include seconds for some reason: https://www.gnu.org/software/trans-coord/manual/gnun/html_node/PO-Header.html

```
"POT-Creation-Date: 2008-02-06 16:25-0500\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
```

### DIFF
--- a/i18n-helpers/src/normalize.rs
+++ b/i18n-helpers/src/normalize.rs
@@ -7,6 +7,7 @@ use std::fs::File;
 use std::io::Read;
 
 use crate::{extract_messages, new_cmark_parser, wrap_sources};
+use chrono::Duration;
 use polib::catalog::Catalog;
 use polib::message::{Message, MessageFlags, MessageMutView, MessageView};
 use pulldown_cmark::{Event, LinkType, Tag};
@@ -247,8 +248,17 @@ pub fn normalize(catalog: Catalog) -> anyhow::Result<Catalog> {
                 message.source_mut().push('\n');
                 message.source_mut().push_str(new_message.source());
             }
-            None => new_catalog.append_or_update(new_message),
+            None => new_catalog.append_or_update(new_message)
         }
+    }
+
+    // Increment POT-Creation-Date in metadata by 1 minute to indicate a change in the file.
+    if let Ok(pot_creation_date) = dateparser::parse(&new_catalog.metadata.pot_creation_date) {
+        new_catalog.metadata.pot_creation_date = pot_creation_date
+            .checked_add_signed(Duration::minutes(1))
+            .ok_or(anyhow::Error::msg("could not advance POT-Creation-Date"))?
+            .format("%Y-%m-%d %H:%M%z")
+            .to_string();
     }
 
     Ok(new_catalog)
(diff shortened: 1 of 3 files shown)


i18n-helpers/src/normalize.rs
```diff
@@ -247,10 +248,19 @@ pub fn normalize(catalog: Catalog) -> anyhow::Result<Catalog> {
...
-            None => new_catalog.append_or_update(new_message),
+            None => new_catalog.append_or_update(new_message)
         }
     }
 
+    // Increment POT-Creation-Date in metadata by 1 minute to indicate a change in the file.
+    if let Ok(pot_creation_date) = dateparser::parse(&new_catalog.metadata.pot_creation_date) {
+        new_catalog.metadata.pot_creation_date = pot_creation_date
+            .checked_add_signed(Duration::minutes(1))
+            .ok_or(anyhow::Error::msg("could not advance POT-Creation-Date"))?
```
mgeisler, line 259: Could you instead leave the date unchanged if it cannot be parsed?

mgeisler (changes requested):
Thanks a lot, this looks great! I just think we should avoid erroring out in case the date cannot be parsed for some unknown reason.

i18n-helpers/src/normalize.rs
```diff
@@ -566,4 +576,12 @@ mod tests {
             ],
         );
     }
+
+    #[test]
+    fn test_normalize_pot_creation_date() {
```
mgeisler, line 581: Please also add a test for the case where the date cannot be parsed.

mgeisler:
> Advances POT-Creation-Date by 1 minute rather than 1 second as @mgeisler suggested as it seems like the date format does not include seconds for some reason: https://www.gnu.org/software/trans-coord/manual/gnun/html_node/PO-Header.html

Nice, I had not noticed that :smile:

i18n-helpers/src/normalize.rs
carreter: Done!

i18n-helpers/src/normalize.rs
carreter: Done!

mgeisler (approved):
Thanks, looks great!

carreter:
@mgeisler Fixed the linting issues, mind giving approval for the workflow?

kdarkhan (approved):

Issue #171, opened 2024-02-06
Normalization does not respect line number granularity


mgeisler:
As a side-quest (:smile:) to #154, I've realized that `mdbook-i18n-normalize` doesn't use the same rounding logic... that will come back and cause problems at some point.

It would be great to have it round the normalized message sources the same way.

zachcmadsen:
I'll pick this up since it's related to #154

mgeisler:
Awesome, thanks @zachcmadsen!

carreter:
Notably `mdbook-xgettext` is also not respecting granularity for subchapters! It currently always includes line numbers when rendering messages in subchapters:

```rust
                let source = ctx.config.book.src.join(&source);
                for (lineno, extracted) in extract_messages(&content) {
                    let msgid = extracted.message;
                    let source = format!("{}:{}", source.display(), lineno);
                    add_message(catalog, &msgid, &source, &extracted.comment);
                }
```

Putting up a fix now.

carreter:
@mgeisler working on this now. How do we want the granularity to be configured? Should it be a command-line flag?

mgeisler:
> @mgeisler working on this now. How do we want the granularity to be configured? Should it be a command-line flag?

Hmm, good question! I guess there are two options:

- a command line flag, nice and simple
- attempt to find an parse a `book.toml` file — that is where the configuration should be written in the first place. We can even call [`Config::from_disk`](https://docs.rs/mdbook/latest/mdbook/config/struct.Config.html#method.from_disk) from `mdbook` to do this.

If it's easy to replicate how `mdbook` loads the config, then can could be nice. If it's not super easy, then a command line flag would be great too.

PR #215, opened 2024-07-19
Make xgettext respect granularity for sub-chapter messages

carreter:
Related to #171.

While trying to refresh the Spanish translation of Comprehensive Rust, @henrif75 noticed that the `messages.pot` file I generated had line numbers that shouldn't be there (see https://github.com/google/comprehensive-rust/issues/2120).

Digging around in the code, I noticed that the `xgettext` code was ignoring the granularity for sub-chapters. This PR fixes that!

### DIFF
--- a/i18n-helpers/src/xgettext.rs
+++ b/i18n-helpers/src/xgettext.rs
@@ -768,7 +768,7 @@ mod tests {
 
         Ok(())
     }
- 
+
     #[test]
     fn test_split_catalog_nested_directories_depth_1() -> anyhow::Result<()> {
         let (ctx, _tmp) = create_render_context(&[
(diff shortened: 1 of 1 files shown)


kdarkhan:
Thanks for the fix, @carreter! Can you add or update one of the existing test cases which will catch this bug please?

I think it will be good to have this tested to ensure we don't break it in the future.

carreter:
Added the test as requested @kdarkhan !

kdarkhan (approved):

kdarkhan:
Please fix formatting. LGTM otherwise.

kdarkhan:
I added the fix on top of your PR, thanks for the contribution.

Issue #220, opened 2024-08-06
Add github-actions check for validating Cargo.lock is up to date


kdarkhan:
More background in https://github.com/google/mdbook-i18n-helpers/pull/216#issuecomment-2255799969

mgeisler:
My own favorite way of fixing this is to automate the releases :smile: 

I've done that for my own _library_ crates using https://github.com/mgeisler/rust-release-gh-action. It is used here, as an example: [`prepare-release.yml`](https://github.com/mgeisler/version-sync/blob/master/.github/workflows/prepare-release.yml), [`publish-crate.yml`](https://github.com/mgeisler/version-sync/blob/master/.github/workflows/publish-crate.yml).

The code doesn't know anything about workspaces, unfortunately, but I would of course like to see it gain that support.

PR #221, opened 2024-08-06
Validate Cargo.lock file is up-to-date

kdarkhan:
Fixes https://github.com/google/mdbook-i18n-helpers/issues/220

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,11 +23,16 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
+      - name: Validate Cargo.lock is up-to-date
+        run: |
+          cargo check --workspace
+          git diff --exit-code
+
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
 
       - name: Run tests with Coverage report enabled
-        run: cargo llvm-cov test --all-features --workspace --codecov --output-path codecov_report.json
+        run: cargo llvm-cov test --all-features --workspace --codecov --locked --output-path codecov_report.json
 
       - name: Store PR number and commit SHA
         run: |


.github/workflows/test.yml
```diff
@@ -176,3 +176,8 @@ jobs:
 
       - name: Check Formatting
         uses: dprint/check@v2.2
+
+      - name: Check Cargo.lock is up-to-date
+        run: |
+          cargo build --workspace
```
kdarkhan, line 182: Not adding `--locked` param here because it will detect only that deps listed in Cargo.lock match what's needed to build the project but won't validate formatting issues such us newlines.

Without `--locked` param this should catch both bad version and formatting issues in `Cargo.lock`. Example of badly formatting `Cargo.lock` [here](https://github.com/google/mdbook-i18n-helpers/actions/runs/10270090217).

qwandor: This is going to take a while, can it be part of the `test` job and/or run `cargo check` rather than `cargo build`?

kdarkhan: `cargo build` takes 45 seconds without using cache and 14 seconds with cache enabled.
`cargo check` is 11 seconds with cache enabled. I am ok with either.

Replaced it with `cargo check` and moved it to `Test` job.

qwandor (approved):

.github/workflows/test.yml
```diff
@@ -23,11 +23,16 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
+      - name: Validate Cargo.lock is up-to-date
+        run: |
+          cargo check --workspace
+          git diff --exit-code
```
mgeisler, line 29: Does this mean that we fail our CI pipelines if a new version has been published to crates.io?

I guess it won't happen as long as the GitHub cache is fresh, but if it expires, then we'll likely see small changes in our `Cargo.lock` file, no?

kdarkhan: Hmm, I haven't used rust enough to know all cases when Cargo.lock is updated.
Based on my understanding, this could fail in 2 cases - PR push or after PR merge.

If during PR push, then the PR creator should re-build the project and Cargo.toml will be updated. That is expected from PR creator I think.

If after PR merge, then this means some of the deps were updated after PR push build but before merge job. I think this is OK as it will be fixed in the next PR.

mgeisler: > Hmm, I haven't used rust enough to know all cases when Cargo.lock is updated.
> Based on my understanding, this could fail in 2 cases - PR push or after PR merge.

I'm afraid of spurious and unrelated changes: I fix a comment and my PR fails because someone released a new patch version of some random crate... That will be confusing for me 🙂

I guess we'll know soon enough and then we can reconsider.

PR #234, opened 2024-10-13
Pulldown cmark to cmark 17.0.0 upgrade

kdarkhan:
* Upgrades pulldown-cmark-to-cmark
* Adjust tests for new way reference links are handled

This supersedes https://github.com/google/mdbook-i18n-helpers/pull/233.

Tests had to be adjusted because new version of the library retains more info from reference links. Reference links are no longer replaced with full URLs. We can probably make some manual adjustments to reference link handling in order to retain the old behavior but I think the new way should work better.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -966,9 +966,9 @@ checksum = "007d8adb5ddab6f8e3f491ac63566a7d5002cc7ed73901f72057943fa71ae1ae"
 
 [[package]]
 name = "pulldown-cmark-to-cmark"
-version = "16.0.1"
+version = "17.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "171b9687038fb417937aee4ce1d0a269897091f05be16660a1e45777533c54b9"
+checksum = "41b27c0d365d60ff3e085007911d73419e5664de48155d558ebfa84d117a5109"
 dependencies = [
  "pulldown-cmark 0.12.1",
 ]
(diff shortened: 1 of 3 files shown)


mgeisler:
> Tests had to be adjusted because new version of the library retains more info from reference links. Reference links are no longer replaced with full URLs.

Oh, that's interesting!

In some sense, having the links expanded is good since it makes the translation completely self-contained. Each `msgid` field is a complete unit. If you change either the text _or_ the link, the whole `msgid` field is invalidated.

However, it has the bad effect of making the text harder to work with for the translators. So I would actually like to see the reference links stay as they are.

I think we need to think about if we can/should keep link definitions together with the `msgid` field that uses them?

What I'm afraid of is a situation where we start with this:

```markdown
Please [click here]!

[click here]: here.html
```

This is translated as:

```po
msgid "Please [click here]!"
msgstr "Venligst [klik her]!"

msgid "[click here]: here.html"
msgstr "[klik her]: her.html"
```

The original text is changed to:

```markdown
Don't [click here]!

[click here]: here.html
```

Because the text changed, `msgmerge` will revert the translation to the English text. The link definition is not changed, so the translation is reused. The translated Markdown document becomes

```markdown
Don't [click here]!

[klik her]: her.html
```

We now have a link without the corresponding link definition!

kdarkhan:
@mgeisler ok, given this I updated the PR to retain the old behavior. I had to edit `expand_shortcut_link` in order to support more link types.

Issue #185, opened 2024-04-18
Support translating using a tree of PO files


mgeisler:
This is the second part of #67: we now support extracting the text into many small PO files. We should thus also support reading these files when translating a book.

hwhsu1231:
How is the progress of this issue?

After the incompatibility issue (https://github.com/google/mdbook-i18n-helpers/issues/278) between `mdbook-i18n-helpers@^0.3` and `mdbook@^0.4` is solved, the primary remaining problem for the [mdbook-docs-l10n](https://github.com/localizethedocs/mdbook-docs-l10n) project is this unfulfilled feature in `mdbook-i18n-helpers`. (though there is a workaround available)

### Problem Description

Although the current `mdbook-i18n-helpers@0.4.0` can generate a tree of POT files using the `depth` config option of the `xgettext` renderer:

```bash
export MDBOOK_OUTPUT='{
  "xgettext": {
    "depth": ${DEPTH}
  }
}'
mdbook build . --dest-dir ./locale/pot
```

It is currently **unable** to build the translated documentation directly using the tree of PO files with the following command:


```bash
export MDBOOK_BOOK__LANGUAGE=${LANGUAGE}
export MDBOOK_PREPROCESSOR='{
  "gettext": {
    "after": ["links"],
    "po-dir": "locale"
  }
}'
mdbook build . --dest-dir book/${LANGUAGE}
```

### Workaround

The current "Workaround" is to manually concatenate the generated tree of PO files into a single `<LANGUAGE>.po` file within the [`mdbook_build_docs`](https://github.com/localizethedocs/mdbook-docs-l10n/blob/954d402ac252b7e9aca1f59064e9b52ae723aa5c/cmake/targets/mdbook_build_docs.cmake#L75-L94) script. This allows the `gettext` preprocessor to read the translation strings from `<LANGUAGE>.po` when building the documentation.


### Detailed Solution

I hope that `mdbook-i18n-helpers` can also add the `depth` config option to the `gettext` preprocessor, which is demonstrated by the following commands:

```bash
DEPTH=5
LANGUAGE=zh_TW

# 1. Generate a tree of .pot files by the `xgettext` renderer.
export MDBOOK_OUTPUT='{
  "xgettext": {
    "depth": ${DEPTH}
  }
}'
mdbook build . --dest-dir locale/pot

# 2. Convert .pot to the .po files with msgcat/msgmerge.
# 
# ```
# locale/
#     pot/
#         summary/
[...]

PR #243, opened 2024-12-13
xgettext: Avoid collisions when splitting .pot files

cip999:
Hey, I've been looking into #185 and its predecessor #67. While thinking of ways to implement the "reconstruction" of the catalog from smaller pieces, I noticed that the code doesn't handle well books where several part titles and/or chapters have the same name (or, more precisely, are equivalent up to `slug`ging).

For example, consider the following `SUMMARY.md`:

```md
# Summary

# Part Foo

[Chapter 1](src/chapter-1.md)

# Part Bar

[Chapter 2](src/chapter-2.md)

# Part Foo

[Chapter 3](src/chapter-3.md)
```

With a `depth` of 1, the preprocessor generates two POT files `foo.pot` and `bar.pot` (beside `summary.pot`), and the contents of the two Part Foo's (including Chapters 1 and 3) are merged. Since those are unrelated, if not for having the same title, the expected behavior would be to put them in separate files. Note that the titles don't have to be exactly equal, for example the same would happen with `Part Foo 1` and `Part Foo 2` (since numbers are discarded).

This PR should fix this issue. I hope doc comments are explicative enough, otherwise feel free to request clarifications. :)


cip999:
Ah, the PR is somwhat large so let me know if you want me to split it into 2-3 commits.

i18n-helpers/src/xgettext.rs
```diff
@@ -178,63 +182,34 @@ where
...
-    let mut summary_destination = match depth {
-        0 => path::PathBuf::from("messages"),
-        1 => path::PathBuf::from("summary"),
-        _ => path::PathBuf::from(&current_top_level).join("summary"),
-    };
-    let _: bool = summary_destination.set_extension("pot");
-    catalogs.insert(summary_destination, catalog);
+    let mut unique_path_builder = UniquePathBuilder::new(depth);
+    unique_path_builder.maybe_push("summary");
+    unique_path_builder.maybe_push("summary");
```
mgeisler, line 187: This looks like a duplicate?

i18n-helpers/src/xgettext.rs
```diff
@@ -257,79 +233,184 @@ struct Chapter {
...
+                frequency_map: HashMap::new(),
+            }
+        }
+
+        fn format_path_with_counter(&self) -> PathBuf {
+            match self.frequency_map.get(&self.current_path) {
+                None | Some(&0) => self.current_path.clone(),
+                Some(&cnt) => {
+                    let mut file_name = self.current_path.file_name().unwrap().to_os_string();
+                    file_name.push(format!("_{}", cnt));
```
mgeisler, line 351: I prefer `-` over `_` in filenames, where possible. So could we do it like this instead:

```suggestion
                    file_name.push(format!("-{cnt}"));
```

i18n-helpers/src/xgettext.rs
```diff
@@ -257,79 +233,184 @@ struct Chapter {
...
-        })
-        .filter(|word| !word.is_empty())
-        .collect::<Vec<_>>()
-        .join("-")
+pub mod paths {
+    use super::POT_EXTENSION;
+    use std::collections::HashMap;
+    use std::path::PathBuf;
+
+    pub struct UniquePathBuilder {
```
mgeisler, line 278: I really like this idea, very elegant!

Two small things: could the `maybe_push` and `maybe_pop` methods be called just `push` and `pop`? I don't think why the `maybe_` prefix is necessary since the push/pop calls have a real effect on the internal state, even if it might not be visible in the resulting string right away.

Another thing, could you attach the big doc string on the `impl` block below to the struct here? That way the struct is nicely document.

A final small thing: please move the inline `paths` module to it's own file.

i18n-helpers/src/xgettext.rs
cip999: Sure, changed!

i18n-helpers/src/xgettext.rs
cip999: All done, thanks for the suggestions. :slightly_smiling_face:

i18n-helpers/src/xgettext.rs
cip999: It's actually intended, to keep the previous behavior: for depth 2, the summary is stored within `summary/summary.pot` (so the first push is for the directory, the second for the file name). For depth 1 instead this results in a top-level `summary.pot`.

kdarkhan:
@cip999, could you rebase this PR now? There are some conflicts.

cip999:
@kdarkhan done!

i18n-helpers/src/xgettext/paths.rs
```diff
@@ -0,0 +1,148 @@
...
+            return; // Exceeded max depth, do nothing.
+        }
+        // Extend current path with new component.
+        self.current_path = self.format_path_with_counter();
+        self.current_path.push(slug(s.as_ref()));
+        // Update frequency of new path.
+        self.frequency_map
+            .entry(self.current_path.clone())
+            .and_modify(|cnt| *cnt += 1)
+            .or_insert(0);
```
kdarkhan, line 81: Values in frequency maps usually represent counts for the respective keys. Having `0` represent a single repetition is a bit confusing I think. Can this be set to `1`? I think it will make the check on line `51` more readable. To maintain the same name offsets, you can use `cnt - 1` in line 52.

i18n-helpers/src/xgettext/paths.rs
```diff
@@ -0,0 +1,148 @@
...
+            current_path: PathBuf::new(),
+            current_depth: 0,
+            max_depth,
+            frequency_map: HashMap::new(),
+        }
+    }
+
+    fn format_path_with_counter(&self) -> PathBuf {
+        match self.frequency_map.get(&self.current_path) {
+            None | Some(&0) => self.current_path.clone(),
```
kdarkhan, line 51: Nit: I think `&` here is redundant.

cip999: Removed.

i18n-helpers/src/xgettext/paths.rs
cip999: Agreed and modified as suggested. Thanks!

kdarkhan (approved):

kdarkhan:
@cip999 thanks for your contribution, sorry for taking so long to respond to the PR, been pretty busy.

cip999:
> @cip999 thanks for your contribution, sorry for taking so long to respond to the PR, been pretty busy.

No problem whatsoever, we're in no rush. :) Thank you for the great review!

PR #244, opened 2024-12-16
Handle errors returned by `cmark_resume_with_options()`

cip999:
Here's the follow-up to https://github.com/Byron/pulldown-cmark-to-cmark/issues/91!


cip999:
Oh, I see now that the version bump was already done in #242 by @kdarkhan. I'll keep my first commit here (for now) since I find the workaround slightly cleaner than what the merged PR does — then I can rebase either way, depending on what you prefer.

i18n-helpers/src/lib.rs
```diff
@@ -652,10 +652,10 @@ pub fn reconstruct_markdown<'a>(
     // Markdown without the padding to remove the effect of these
     // structural elements. Similarly, we don't want extra newlines at
     // the start.
-    let simplified_state = state.map(|state| State {
-        newlines_before_start: 0,
-        padding: Vec::new(),
-        ..state
+    let simplified_state = state.map(|mut state| {
```
kdarkhan, line 654: This does indeed look cleaner. I did not think that I could work on the state as a mutable within map even though the option itself is immutable. This part looks good to me.

i18n-helpers/src/gettext.rs
```diff
@@ -38,11 +40,21 @@ fn strip_formatting(text: &str) -> String {
...
 
-fn translate(text: &str, catalog: &Catalog) -> String {
+fn translate(text: &str, catalog: &Catalog) -> anyhow::Result<String> {
     let events = extract_events(text, None);
-    let translated_events = translate_events(&events, catalog);
-    let (translated, _) = reconstruct_markdown(&translated_events, None);
-    translated
+    // Translation should always succeed.
+    let translated_events =
+        translate_events(&events, catalog).expect("Fatal: failed to translate events");
```
kdarkhan, line 47: Is this `Fatal:` prefix needed here? It seems to be inconsistent with the rest of `expect()` failures we have in this codebase.

i18n-helpers/src/lib.rs
```diff
@@ -462,7 +462,7 @@ fn heuristic_codeblock<'a>(
         _ => true,
     };
 
-    if is_translate {
+    Ok(if is_translate {
```
kdarkhan, line 465: Nit: wrapping the expression in `Ok` looks to me a bit harder to read but maybe this is the way to do it in rust (haven't done much rust yet). Will this be more readable if you store the result into two variables - `events` and `ctx`, then at the end of the function do `Ok(events, ctx)`?

kdarkhan:
> Oh, I see now that the version bump was already done in #242 by @kdarkhan. I'll keep my first commit here (for now) since I find the workaround slightly cleaner than what the merged PR does — then I can rebase either way, depending on what you prefer.

Thanks for handling this, @cip999! Please rebase this on top of latest changes. Would be good if @mgeisler could take a look as well since he has some context on this issue.

i18n-helpers/src/lib.rs
cip999: Good, I kept it! :)

i18n-helpers/src/gettext.rs
cip999: Right I can't see it anywhere else, removed it. And panic is already fatal anyway...

i18n-helpers/src/lib.rs
cip999: Sure, done! I've written even less rust, so I have no idea what the proper "rust style" would be (or if it there's a preferred way in the first place). But your suggestion looks more readable.

cip999:
> > Oh, I see now that the version bump was already done in #242 by @kdarkhan. I'll keep my first commit here (for now) since I find the workaround slightly cleaner than what the merged PR does — then I can rebase either way, depending on what you prefer.
> 
> Thanks for handling this, @cip999! Please rebase this on top of latest changes. Would be good if @mgeisler could take a look as well since he has some context on this issue.

Thanks for reviewing! Yes that's a good idea, and it can wait, there's no rush to merge this PR right now. :slightly_smiling_face:

kdarkhan (approved):

i18n-helpers/src/gettext.rs
```diff
@@ -108,6 +130,14 @@ mod tests {
         catalog
     }
 
+    fn create_book(items: Vec<BookItem>) -> Book {
+        let mut book = Book::new();
+        items.into_iter().for_each(|item| {
+            book.push_item(item);
+        });
```
mgeisler, line 137: I think a normal for-loop should work here and I think that would be a little simpler:

```suggestion
        for item in items {
            book.push_item(item);
        }
```

cip999: Yeah guess I was too eager on doing things "the rust way" :D

kdarkhan:
Thanks for contribution, @cip999!

PR #253, opened 2025-02-08
Fix broken tables whose msgstr is a specific type of HTML

### LINKED ISSUE #251: Broken table with `<img>` tag
The table with `` tag becomes broken by translation.

* Source

```markdown
| Ferris                                                                                                           | Meaning                                          |
| ---------------------------------------------------------------------------------------------------------------- | ------------------------------------------------ |
| <img src="img/ferris/does_not_compile.svg" class="ferris-explain" alt="Ferris with a question mark"/>            | This code does not compile!                      |
```

* PO

```po
#: src/chapter_1.md:5
msgid ""
"<img src=\"img/ferris/does_not_compile.svg\" class=\"ferris-explain\" alt="
"\"Ferris with a question mark\"/>"
msgstr ""
"<img src=\"img/ferris/does_not_compile.svg\" class=\"ferris-explain\" alt="
"\"Ferris with a question mark\"/>"

#: src/chapter_1.md:5
msgid "This code does not compile!"
msgstr "このコードはコンパイルできません!"
```

* Generated HTML

```html
                        <div class="table-wrapper"><table><thead><tr><th>Ferris</th><th>Meaning</th></tr></thead><tbody>
<tr><td><img src="img/ferris/does_not_compile.svg" class="ferris-explain" alt="Ferris with a question mark"/></td><td></td></tr>
<tr><td>このコードはコンパイルできません!</td><td></td></tr>
</tbody></table>
</div>
```

* Original HTML

```html
                        <div class="table-wrapper"><table><thead><tr><th>Ferris</th><th>Meaning</th></tr></thead><tbody>
<tr><td><img src="img/ferris/does_not_compile.svg" class="ferris-explain" alt="Ferris with a question mark"/></td><td>This code does not compile!</td></tr>
</tbody></table>
</div>
```

According to comparison with the original HTML, extra `<tr>` is appended, and there is empty `<td></td>` where the translated text should be inserted to.

shinmili:
If an msgstr for a table cell contains HTML that could be parsed as an HTML block if it were put in the top-level, the table is rendered incorrectly. The table row containing the cell gets interrupted in the middle and split into two (or more, if there are more such cells) incomplete rows.

Close #251.

### DIFF
--- a/i18n-helpers/src/gettext.rs
+++ b/i18n-helpers/src/gettext.rs
@@ -322,6 +322,34 @@ mod tests {
     }
 
     #[test]
+    fn test_translate_html_in_table() {
+        let catalog = create_catalog(&[
+            ("Icon", "ICON"),
+            ("Description", "DESCRIPTION"),
+            (
+                "<img src=\"some-icon.svg\" alt=\"Some icon.\"/>",
+                "<img src=\"some-icon.svg\" alt=\"SOME ICON.\"/>",
+            ),
+            ("Some description.", "SOME DESCRIPTION."),
+        ]);
+        // The alignment is lost when we generate new Markdown.
+        assert_eq!(
+            translate(
+                "\
+                | Icon                                            | Description       |\n\
+                |-------------------------------------------------|-------------------|\n\
+                | <img src=\"some-icon.svg\" alt=\"Some icon.\"/> | Some description. |",
+                &catalog
+            )
+            .unwrap(),
+            "\
+            |ICON|DESCRIPTION|\n\
+            |----|-----------|\n\
+            |<img src=\"some-icon.svg\" alt=\"SOME ICON.\"/>|SOME DESCRIPTION.|"
+        );
+    }
+
+    #[test]
     fn test_footnote() {
         let catalog = create_catalog(&[
             ("A footnote[^note].", "A FOOTNOTE[^note]."),
--- a/i18n-helpers/src/gettext.rs
+++ b/i18n-helpers/src/gettext.rs
@@ -347,6 +347,24 @@ mod tests {
     }
 
     #[test]
+    fn test_translate_escaped_pipe_in_table() {
+        let catalog = create_catalog(&[("foo\\|bar", "FOO\\|BAR")]);
+        // The alignment is lost when we generate new Markdown.
+        assert_eq!(
+            translate(
+                "\
+                |foo\\|bar|\n\
+                |---------|",
+                &catalog
+            )
+            .unwrap(),
+            "\
+            |FOO\\|BAR|\n\
+            |-------|",
+        );
+    }
+
+    #[test]
     fn test_footnote() {
         let catalog = create_catalog(&[
             ("A footnote[^note].", "A FOOTNOTE[^note]."),
--- a/i18n-helpers/src/gettext.rs
+++ b/i18n-helpers/src/gettext.rs
@@ -326,26 +326,23 @@ mod tests {
         let catalog = create_catalog(&[
             ("Icon", "ICON"),
             ("Description", "DESCRIPTION"),
-            (
-                "<img src=\"some-icon.svg\" alt=\"Some icon.\"/>",
-                "<img src=\"some-icon.svg\" alt=\"SOME ICON.\"/>",
-            ),
+            ("<img src=\"some-icon.svg\"", "<img src=\"some-icon.svg\""),
             ("Some description.", "SOME DESCRIPTION."),
         ]);
         // The alignment is lost when we generate new Markdown.
         assert_eq!(
             translate(
                 "\
-                | Icon                                            | Description       |\n\
-                |-------------------------------------------------|-------------------|\n\
-                | <img src=\"some-icon.svg\" alt=\"Some icon.\"/> | Some description. |",
+                | Icon                         | Description       |\n\
+                |------------------------------|-------------------|\n\
+                | <img src=\"some-icon.svg\"/> | Some description. |",
                 &catalog
             )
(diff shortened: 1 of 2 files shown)


i18n-helpers/src/lib.rs
```diff
@@ -137,6 +140,33 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'a>>) -> Vec<(usize
             .enumerate()
             .map(|(idx, line)| (idx + 1, Event::Text(line.into())))
             .collect(),
+        // If we're in a table cell, we put the text in a minimal table, parse the
+        // table, and return the contents of the cell. This matches the behavior of
+        // the parser in this case.
+        Some(state) if state.in_table_cell => {
+            let text = format!("|{}|\n|-|", text);
```
mgeisler, line 147: I'm fond of the inline formatting directives and use them where possible:

```suggestion
            let text = format!("|{text}|\n|-|");
```

i18n-helpers/src/lib.rs
```diff
@@ -137,6 +140,33 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'a>>) -> Vec<(usize
...
+                        Event::Start(Tag::Table(..) | Tag::TableHead | Tag::TableCell)
+                        | Event::End(TagEnd::Table | TagEnd::TableHead | TagEnd::TableCell) => {
+                            return None
+                        }
+                        Event::SoftBreak => Event::Text(" ".into()),
+                        // Shortcut links like "[foo]" end up as "[foo]"
+                        // in output. By changing them to a reference
+                        // link, the link is expanded on the fly and the
+                        // output becomes self-contained.
+                        Event::Start(tag @ (Tag::Link { .. } | Tag::Image { .. })) => {
```
mgeisler, line 160: This is duplicating some existing functionality, right? I feel I remember seeing this before :smile: 

If so, we should pull this filtering code out into its own function so that we can call it from here as well.

i18n-helpers/src/lib.rs
```diff
@@ -137,6 +140,33 @@ pub fn extract_events<'a>(text: &'a str, state: Option<State<'a>>) -> Vec<(usize
             .enumerate()
             .map(|(idx, line)| (idx + 1, Event::Text(line.into())))
             .collect(),
+        // If we're in a table cell, we put the text in a minimal table, parse the
```
mgeisler, line 143: A question about the minimal table: what happens if the text contains a `|` character?

I am not 100% sure what the Commonmark rules are for escaping pipes in tables — can you add another test to document the behavior if it's supported? If it's not supported, I would appreciate a comment here saying that we're sure `text` doensn't contain `|` because we know we're in a table context.

i18n-helpers/src/gettext.rs
```diff
@@ -321,6 +321,34 @@ mod tests {
...
+            ),
+            ("Some description.", "SOME DESCRIPTION."),
+        ]);
+        // The alignment is lost when we generate new Markdown.
+        assert_eq!(
+            translate(
+                "\
+                | Icon                                            | Description       |\n\
+                |-------------------------------------------------|-------------------|\n\
+                | <img src=\"some-icon.svg\" alt=\"Some icon.\"/> | Some description. |",
```
mgeisler, line 341: Just an idea: consider "translating the `src` attribute and dropping the `alt` attribute to make the example smaller. I believe one attribute should be enough to demonstrate this.

mgeisler:
Hi @shinmili, thank you so much for putting this up! I think this looks good and I left some minor comments.

i18n-helpers/src/lib.rs
shinmili: Right, some parts are copied from the `_ =>` arm below. I'll try to making such function.

i18n-helpers/src/gettext.rs
shinmili: Sounds good! I just intended to write some practical example, but it isn't really necessary for testing.

i18n-helpers/src/lib.rs
shinmili: I would say it's basically supported, so I'll add a test for that.

Technically, I can't say `text` doesn't contain `|`, because translators could write arbitrary `msgstr`s in `.po`... As long as they don't introduce any non-escaped `|`s, it works fine.

mgeisler: Thanks a lot! 

> As long as they don't introduce any non-escaped `|`s, it works fine.

That is great! The policy is basically: if people put garbage in the `.po` files, then it's okay to output garbage :smile:

mgeisler (approved):

i18n-helpers/src/lib.rs
mgeisler: Thanks for the function! However, now Clippy is unhappy with the code because of unnecessary lifetimes:

```
error: the following explicit lifetimes could be elided: 'a
   --> i18n-helpers/src/lib.rs:129:29
    |
129 |     fn convert_event_common<'a>(event: Event<'a>) -> Event<'a> {
    |                             ^^               ^^            ^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
    = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_lifetimes)]`
help: elide the lifetimes
    |
129 -     fn convert_event_common<'a>(event: Event<'a>) -> Event<'a> {
129 +     fn convert_event_common(event: Event<'_>) -> Event<'_> {
    |
```

Would you be able to take this out?

Issue #235, opened 2024-10-29
Bug: Some scenes were found to be untranslatable into the corresponding Chinese.


lispking:
## **po file**: 
https://github.com/lispking/walrus-docs/blob/main/po/zh_CN.po#L251
![Image](https://github.com/user-attachments/assets/7420d90c-4891-4661-a620-d3e1f2c4881a)

## **Effect**:
![Image](https://github.com/user-attachments/assets/11577811-2239-4d9f-a56b-0b1ad11aee4a)

mgeisler:
Hi @lispking, thanks for opening the issue!

Okay, this looks strange indeed and I'm not sure why this wouldn't work. I guess you can confirm that your `zh_CN.po` file has an up-to-date `msgid` for this text?

I think we could use a debug command which can help show us the text we find in the Markdown files. That would help us track down if we're looking up the right message in the `.po` file.

lispking:
Hello @mgeisler , There is a corresponding msgid.

I don't know if it's related to the following line.

```admonish tip title=\"Fun fact\"\n"

mgeisler:
Yes, it might be related to "info string" after the ` ``` `. We are supposed to pass those though unchanged, but we might have a bug there, I'm not sure 😄

I won't have time to debug this myself right now, but I hope I can find someone with some free time to look at this in the coming weeks.

PR #264, opened 2025-04-20
Fix translation in custom code block

jamielapointe:
Fix [issue#235](https://github.com/google/mdbook-i18n-helpers/issues/235).

This adds a simple basic handler for translating [mdbook-admonish](https://github.com/tommilligan/mdbook-admonish) custom code blocks.  The code specifically looks for the admonish handler in the code block and does a simple block translation, rather than trying to do complex parsing.

To use this, you would do something like the following your `book.toml`:

```toml
[preprocessor.admonish]
command = "mdbook-admonish"
on_failure = "bail"
assets_version = "3.0.2" # do not edit: managed by `mdbook-admonish install`

#...

[preprocessor.gettext]
before = ["admonish"]
after = ["links"]
```

An example *.po would look something like this:

```bash
msgid ""
"```admonish tip title=\"Debug logging\"\n"
"You can enable debug logging for Walrus by setting the environment variable "
"`RUST_LOG=walrus=debug`.\n"
"```"
msgstr ""
"```admonish tip title=\"调试日志记录\"\n"
"您可以通过设置环境变量 `RUST_LOG=walrus=debug` 来启用 Walrus 的调试日志记"
"录。\n"
"```"
```


i18n-helpers/src/lib.rs
```diff
@@ -493,8 +527,25 @@ fn parse_codeblock<'a>(
...
-        return heuristic_codeblock(events, ctx);
+        // This is to handle the special case mdbook-admonish custom
+        // "code block"
+        // https://github.com/tommilligan/mdbook-admonish
+        static ADMONISH_CODEBLOCK_NAME: &str = "admonish";
+        let is_admonish =
+            if let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(x)))) = &events[0] {
+                x.split(',')
+                    .next()
+                    .unwrap()
```
mgeisler, line 538: Please avoid the `unwrap` since it can panic (crash!). Since you do a `contains` afterwards, you can use `unwrap_or_default` to instead get an empty string (the default value). That should give the same behavior, while explicitly ensuring no crash.

i18n-helpers/src/lib.rs
```diff
@@ -493,8 +527,25 @@ fn parse_codeblock<'a>(
     };
 
     let Some(syntax) = syntax else {
-        // If there is no language specifier, falling back to heuristic way.
-        return heuristic_codeblock(events, ctx);
+        // This is to handle the special case mdbook-admonish custom
+        // "code block"
+        // https://github.com/tommilligan/mdbook-admonish
+        static ADMONISH_CODEBLOCK_NAME: &str = "admonish";
```
mgeisler, line 533: I think `const` instead of `static` would be simpler. That is, as a Rust developer, I care about this being a constant — I don't (typically) care if it has a static memory location.

i18n-helpers/src/lib.rs
```diff
@@ -477,6 +477,40 @@ fn heuristic_codeblock<'a>(
...
+/// If a title is present, also extract that for translation
+/// Note this is from mdbook-admonish: https://github.com/tommilligan/mdbook-admonish
+fn admonish_codeblock<'a>(
+    events: &'a [(usize, Event<'_>)],
+    mut ctx: GroupingContext,
+) -> Result<(Vec<Group<'a>>, GroupingContext), CmarkError> {
+    // First, ensure we're dealing with a code block
+    if events.len() < 2
+        || !matches!(&events[0].1, Event::Start(Tag::CodeBlock(_)))
+        || !matches!(&events[events.len() - 1].1, Event::End(TagEnd::CodeBlock))
```
mgeisler, line 490: Please try to turn this into a [slice pattern](https://doc.rust-lang.org/reference/patterns.html#slice-patterns):

```rust
match events {
  [Event::Start(Tag::CodeBlock(_), .., Event::Start(Tag::CodeBlock(_)] => // do the thing
  _ => // error out
}
```

i18n-helpers/src/lib.rs
```diff
@@ -477,6 +477,40 @@ fn heuristic_codeblock<'a>(
...
+    {
+        return heuristic_codeblock(events, ctx);
+    }
+
+    // Check if we're dealing with a fenced code block
+    if !matches!(
```
mgeisler, line 496: The `matches!` macro is not typically your first tool. Constructs like `match`, `if let`, and the new `let patter = expression else { return ... }` are your friend.

i18n-helpers/src/lib.rs
jamielapointe: Yes, very good, that was an oversight... in C++ I would normally make constants like this `static const`, that was probably my thinking.

i18n-helpers/src/lib.rs
jamielapointe: Another oversite, thanks for catching this

i18n-helpers/src/lib.rs
```diff
@@ -493,8 +527,25 @@ fn parse_codeblock<'a>(
...
+        let is_admonish =
+            if let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(x)))) = &events[0] {
+                x.split(',')
+                    .next()
+                    .unwrap()
+                    .contains(ADMONISH_CODEBLOCK_NAME)
+            } else {
+                false
+            };
+        if is_admonish {
```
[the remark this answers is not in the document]
jamielapointe: Yes, agreed

i18n-helpers/src/lib.rs
jamielapointe: It felt wrong to add identical checks to the `admonish_codeblock` function that were previously already checked in the calling function.  I added some additional notes into the comments; Please let me know if this acceptable or not.

i18n-helpers/USAGE.md
```diff
@@ -155,6 +155,23 @@ after = ["links"]
...
+`book.toml` file like so:
+
+```toml
+[preprocessor.admonish]
+command = "mdbook-admonish"
+on_failure = "bail"
+assets_version = "3.0.2" # do not edit: managed by `mdbook-admonish install`
+
+#...
+
```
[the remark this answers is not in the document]
jamielapointe: Sounds good, I agree.

i18n-helpers/src/lib.rs
```diff
@@ -1796,4 +1848,115 @@ print("Hello world")
...
+    #[test]
+    fn test_admonish_codeblock_extraction_empty() {
+        // Empty events should not produce any messages
+        assert_extract_messages("", &[]);
+    }
+
+    #[test]
+    fn test_admonish_codeblock_extraction_not_code_block() {
+        // Regular text should be extracted normally, not as admonish
+        assert_extract_messages(
```
mgeisler, line 1918: This is not really about `admonish_codeblock` — do we not have other coverage of `assert_extract_messages`?

i18n-helpers/src/lib.rs
```diff
@@ -477,6 +477,54 @@ fn heuristic_codeblock<'a>(
...
+    let (_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(info_string)))) = &events[0] else {
+        return false;
+    };
+
+    // Check if the language specifier contains "admonish"
+    info_string
+        .split(',')
+        .next()
+        .unwrap_or_default()
+        .contains(ADMONISH_CODEBLOCK_NAME)
```
mgeisler, line 525: Would this do more or less the same:

```suggestion
    match events {
        [(_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(info_string)))), .., (_, Event::End(TagEnd::CodeBlock))] => {
            info_string.split(' ').next() == Some(ADMONISH_CODEBLOCK_NAME)
        }
        _ => false,
    }
```

This changes the splitting to use `' '` and uses a slice pattern to assert the overall structure in one go.

I skimmed the `mdbook-admonish` code and found [this code](https://github.com/tommilligan/mdbook-admonish/blob/21d86546ffbcad83108f163c5805737001c0c2a3/src/config/mod.rs#L19-L32), which seems to split on `' '` instead.

Down the road, I think we can avoid special-casing `admonish` and instead have a configurable list of languages (from the info string) which should be fully extracted. We can also have a list of strings that should be unconditionally ignored. But that is for another PR :smile:

i18n-helpers/src/gettext.rs
```diff
@@ -276,6 +276,146 @@ mod tests {
...
+            )
+            .unwrap(),
+            "```admonish tip\n\
+            星波轻颤动，\n\
+            中子核密语声声，\n\
+            奥秘渐分明。\n\
+            ```",
+        );
+    }
+
```
jamielapointe, line 418: I removed the redundant test - not sure why I had two identical tests - but I finally noticed it and removed it.

i18n-helpers/src/lib.rs
```diff
@@ -477,6 +477,41 @@ fn heuristic_codeblock<'a>(
...
+}
+
+/// Check if the code block is an admonish block
+fn is_admonish(events: &[(usize, Event<'_>)]) -> bool {
+    const ADMONISH_CODEBLOCK_NAME: &str = "admonish";
+
+    // pull the info_string (aka language specifier) out of the code block
+    // (the string after the ```)
+    match events {
+        [(_, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(info_string)))), .., (_, Event::End(TagEnd::CodeBlock))] =>
```
jamielapointe, line 506: Is this more acceptable?  I agree it is much cleaner.  I am using `split_once(' ')` which splits at the first occurrence of the split char (an empty space) which is consistent with how admonish was doing it, as you point out.  Please let me know if there is a safer/more robust way to do this.

I agree with your previous comments that we eventually do not need a special "admonish" handler... as it is not really parsing anything special out of the block.

i18n-helpers/src/lib.rs
```diff
@@ -1796,4 +1835,67 @@ print("Hello world")
             ),]
         );
     }
+
```
jamielapointe, line 1838: I took your advice on the tests.  It was a bit ridiculous and did not really match how the other unit tests were setup.  I think this is better and a bit cleaner.  I was trying to set up a test to test the default handler of the match in the `is_admonish` function, but I just could not do it using `assert_extract_messages`, probably because of the checks prior to getting there - ensuring we are in a code block and such without calling it directly.  However, as you mentioned, we probably do not want to call it directly with manually manipulated Events array.  I was attempting to improve the code coverage of the new code I added.

jamielapointe: If I had to test the default handler of the match events in `is_admonish` just to ensure it is reachable and does not break anything, then it would look something like this; however, as you have mentioned previously, it is unnecessary and does not adhere the style of the other tests; in addition we probably should not strive for 100% code coverage just for code coverage's sake.

```rust
    // This will test the we successfully hit the default handler of the match
    // events in `is_admonish`
    #[test]
    fn is_admonish_codeblock_no_end_block() {
        let events = vec![(
            1,
            Start(CodeBlock(CodeBlockKind::Fenced("admonish tip ".into()))),
        )];
        assert!(!is_admonish(&events));
    }
```

i18n-helpers/src/gettext.rs
mgeisler: Thank you!

mgeisler (approved):
Thank you, this looks great! Let's merge and release this :smile:

PR #265, opened 2025-04-30
Make syntect fancy

cuviper:
This switches `syntect` from `onig` to `fancy-regex` for a pure Rust
implementation of regular expressions.

It's also worth noting that `onig`'s underlying `oniguruma` library says
"This project ended on April 24, 2025", with its [repo archived][1].

[1]: https://github.com/kkos/oniguruma/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,6 +112,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


cuviper:
I don't know that I care to deal with a CLA for such a small change. I welcome anyone else to recreate this, which should be trivial even without looking at my commit.

mgeisler:
> I don't know that I care to deal with a CLA for such a small change. I welcome anyone else to recreate this, which should be trivial even without looking at my commit.

Yeah, I understand completely! Sorry for the speed bump, it's a Google-wide policy to require an explicit signoff via the CLA.

I'll recreate it like you suggest. Thanks for the PR!

mgeisler:
#267 should do the same, thanks again!

PR #268, opened 2025-05-12
Build and update script for translations

### LINKED ISSUE #18: Add scripts for using and updating translations

jonodeno:
Adds a script with build and update sub-commands 

Fixes #18.

### DIFF
--- a/i18n-helpers/src/bin/mdbook-translate-helper.rs
+++ b/i18n-helpers/src/bin/mdbook-translate-helper.rs
@@ -67,7 +67,6 @@ fn build_translation(locale: String, dest_dir: String) -> Result<(), Error>{
 
   // Disable the redbox button in built versions of the course
   fs::write(format!("{}/html/theme/redbox.js", dest_dir), "// Disabled in published builds, see build.sh")?;
-
   let pdf_from_dir = format!("{}/pandoc/pdf/comprehensive-rust.pdf", dest_dir);
   let pdf_dest_dir = format!("{}/html/comprehensive-rust.pdf", dest_dir);
 
@@ -117,11 +116,12 @@ fn update() {
   unimplemented!("update has not been implemented yet")
 }
 
-fn build(dest_dir: String) -> Result<(), Error>{
+fn build(dir: &str) -> Result<(), Error>{
   match env::var("LANGUAGES") {
     Ok(languages) => {
       for language in languages.split(" ") {
-        match build_translation(language.to_string(), dest_dir.clone()) {
+        let dest_dir = dir.to_owned() + "/" + language;
+        match build_translation(language.to_string(), dest_dir) {
           Ok(_) => (),
           Err(err) => return Err(err.into())
         }
@@ -147,7 +147,8 @@ fn run_helper(args: &[String]) -> () {
       }
 
       if translate_action.as_str() == "build" {
-        match build(args[2].clone()) {
+        let dir = args[2].clone();
+        match build(&dir) {
           Ok(_) => (),
           Err(err) => panic!("::endgroup::Error building translations: {}", err)
         }
--- a/i18n-helpers/src/bin/mdbook-translate-helper.rs
+++ b/i18n-helpers/src/bin/mdbook-translate-helper.rs
@@ -115,8 +115,18 @@ fn build_translation(locale: String, dest_dir: String) -> Result<(), Error>{
 }
 
 
-fn update() {
-  unimplemented!("update has not been implemented yet")
+fn update(filename: &str) -> Result<(), Error> {
+  match Command::new("mdbook").arg("build").output() {
+    Ok(_) => (),
+    Err(err) => return Err(err.into())
+  };
+
+  match Command::new("msgmerge").args(["--update", filename, "book/xgettext/messages.pot"]).output() {
+    Ok(_) => (),
+    Err(err) => return Err(err.into())
+  };
+
+  Ok(())
 }
 
 fn build(dir: &str) -> Result<(), Error>{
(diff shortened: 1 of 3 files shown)


i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+          Err(err) => return Err(err.into())
+        }
+      }
+      return Ok(())
+    },
+    Err(err) => return Err(err.into())
+  }
+}
+
+fn run_helper(args: &[String]) -> () {
```
michael-kerscher, line 148: We could leverage [clap](http://docs.rs/clap) for parsing the arguments here

i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+
+  match Command::new("msgmerge").args(["--update", filename, "book/xgettext/messages.pot"]).output() {
+    Ok(_) => (),
+    Err(err) => return Err(err.into())
+  };
+
+  Ok(())
+}
+
+fn build(dir: &str) -> Result<(), Error>{
```
michael-kerscher, line 132: dir could be a [Path](https://doc.rust-lang.org/std/path/struct.Path.html). That type has some nice operators and this resolves the issue in line 136 with manual path separators (Windows is not so keen on '/' ;))

i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+use std::process::Command;
+use anyhow::Context;
+use anyhow::Error;
+use chrono;
+use walkdir::WalkDir;
+use zip::result::ZipError;
+use zip::write::SimpleFileOptions;
+use std::fs::File;
+
+fn build_translation(locale: String, dest_dir: String) -> Result<(), Error>{
```
michael-kerscher, line 17: May I ask you to split up that function into smaller functions to better spot the individual steps of the build?

michael-kerscher (commented):
Hi Jonathan. Thank you for this PR, this looks great and useful!

The CI piplines spotted some issues that need to be fixed.

I also gave it a first pass and have some small comments:
I noticed, that there are a lot of path "Strings" that from my perspective would benefit in being a [Path](https://doc.rust-lang.org/std/path/struct.Path.html). This resolves a lot of issues with Windows vs. Linux (directory separators...) and has good ergonomics that don't require a lot of string formatting :)

Regarding argument parsing, we already use clap in this repository that removes the burden of parsing and verifying the command line arguments from your shoulders :)
For an example in this repository you can have a look at https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-report/src/main.rs#L53-L69 - this example also uses Paths and defaults.

i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+  if locale == "en" {
+    println!("::group::Building English course");
+  } else {
+    let locale_ref = &locale;
```
mgeisler, line 21: We don't need a separate variable for this -- `&locale` is shorter and more clear to me :)

i18n-helpers/src/bin/mdbook-translate-helper.rs
mgeisler, line 17: Functions very rarely need to take an owned `String` as an argument. Normally you want your functions to take `&str` instead.

The reason is that this gives the caller more flexibility: they can do both

```rust
let s = String::from("this is a heap-allocated owned string");
build_translation(&s);
```

and

```rust
build_translation("static non-heap allocated string");
```

When you take a `String` in your function signature, you force the caller to allocate memory for it.

i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+        break;
+      }
+    }
+
+    if pot_creation_date.is_none() {
+      let now = chrono::Local::now();
+      pot_creation_date = Some(now.format("%Y-%m-%dT%H:%M:%S").to_string());
+    }
+
+    println!("::group::Building {locale_ref} translation as of {:?}", pot_creation_date.clone().unwrap());
```
mgeisler, line 41: Every `unwrap()` is a potential for a crash, so you should avoid them in production code.

Here, it looks like you initialize `pot_creation_date` to `Some(...)` in any case? If so, the type should not be `Option<String>` but just `String`. The unwrap here is basically you working around the type being inconvenient for you.

What you can do, is to initialize it like you do in the loop, and then afterwards use [`Option::unwrap_or_else`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unwrap_or_else) to get a default value with `chrono::Local::now()`.

The result would be a `String`, not an `Option<String>`.

i18n-helpers/src/bin/mdbook-translate-helper.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    // Back-date the source to POT-Creation-Date. The content lives in two directories:
+    fs::remove_dir_all("src")?;
+    fs::remove_dir_all("third_party")?;
+
+    let output = Command::new("git").args(["rev-list", "-n", "1", "--before", &pot_creation_date.unwrap(), "@"]).output().unwrap();
+    let result_str= String::from_utf8(output.stdout).unwrap().replace("\n", "");
+
+    Command::new("git").args(["restore","--source", &result_str, "src/", "third_party/"]).output()?;
```
mgeisler, line 50: This line needs to take https://github.com/google/comprehensive-rust/pull/2658 into account :smile:

In general, please extract the paths to restore to a `const` at the global level. That way we can maintain it better.

jonodeno: I tried `Command::new("git").args(["restore","--source", &result_str, "src/", "third_party/"]).output()?;` and even when I run `git restore --source "$(git rev-list -n 1 --before "$pot_creation_date" @)" src/ third_party/ book.toml` manually, pandoc fails and the comprehensive-rust pdf isn't created. 

I just get something like:
```
~/comprehensive-rust$ mdbook build -d book/fa
2025-07-15 22:29:25 [INFO] (mdbook::book): Book building has started
2025-07-15 22:29:26 [INFO] (mdbook::book): Running the exerciser backend
2025-07-15 22:29:26 [INFO] (mdbook::renderer): Invoking the "exerciser" renderer
2025-07-15 22:29:27 [INFO] (mdbook::book): Running the html backend
2025-07-15 22:29:28 [INFO] (mdbook::book): Running the linkcheck backend
2025-07-15 22:29:28 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
2025-07-15 22:29:28 [INFO] (mdbook::book): Running the pandoc backend
2025-07-15 22:29:28 [INFO] (mdbook::renderer): Invoking the "pandoc" renderer
2025-07-15 22:29:29 [INFO] (mdbook_pandoc::pandoc::renderer): Running pandoc
Error producing PDF.
!  ==> Fatal error occurred, no output PDF file produced!

pandoc exited unsuccessfully
2025-07-15 22:29:42 [ERROR] (mdbook::renderer): Renderer exited with non-zero return code.
2025-07-15 22:29:42 [ERROR] (mdbook::utils): Error: Rendering failed
2025-07-15 22:29:42 [ERROR] (mdbook::utils):    Caused By: The "pandoc" renderer failed
```

Could it be some sort of mismatch between the restored book.toml and the recent update to Rust 2024 edition?

PR #296, opened 2026-02-14
Allow translating HTML blocks

### LINKED ISSUE #254: Add support for translating HTML blocks
The current implementation of `mdbook-xgettext` extracts (inline) raw HTML tags as translatable, but skips HTML blocks. So the extracted .po file has msgid for this `img`:

```markdown
...

<img alt="Two tables: the first table contains the representation of s1 on the
stack, consisting of its length (5), capacity (5), and a pointer to the first
value in the second table. The second table contains the representation of the
string data on the heap, byte by byte." src="img/trpl04-01.svg" class="center"
style="width: 50%;" />

...
```

but not for this `img`:

```markdown
...

<img alt="Rendered HTML documentation for the `add_one` function of `my_crate`" src="img/trpl14-01.png" class="center" />

...
```

I want the latter one to be translatable too, especially for `alt` texts. Why does it skip HTML blocks?

shinmili:
Fix #254.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -395,20 +395,9 @@ pub fn group_events<'a>(events: &'a [(usize, Event<'a>)]) -> Result<Vec<Group<'a
                     }
                     _ => {
                         match event {
-                            Event::Html(_) => {
-                                // Otherwise, treat as a skipping group if this is a block level Html tag
-                                if let State::Translate(_) = state {
-                                    let mut next_groups;
-                                    (next_groups, ctx) = state.into_groups(idx, events, ctx)?;
-                                    groups.append(&mut next_groups);
-
-                                    state = State::Skip(idx);
-                                }
-                            }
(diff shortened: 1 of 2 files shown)


qwandor (approved):
Seems reasonable to me, but I'm not very familiar with this.

i18n-helpers/src/lib.rs
```diff
@@ -1457,7 +1454,9 @@ HTML -->
              Postamble",
             &[
                 (1, "Preamble"), //
+                (3, "<details>\n"),
                 (5, "Some Details"),
+                (7, "</details>\n"),
                 (9, "Postamble"),
```
mgeisler, line 1460: Super, this is the important part: we use `<details>` a lot in Comprehensive Rust, and this change shows that our existing translations will survive: the `<details>\n` and `</details>\n` lines will simply become two new untranslated messages. Which is no big deal.

mgeisler (approved):

PR #310, opened 2026-06-25
fix char-boundary panic in comment directive parsing

sage-mode-hunter:
1. The comment directive parser locates the `comment` keyword and skips the keyword length plus one byte to reach the body, assuming the separator after the keyword is a single byte.
2. The tokenizer treats any Unicode whitespace as a separator, so a multi-byte one like U+00A0 makes that byte offset land inside the code point, and `directives::find` panics slicing on a non-char boundary for book content such as `<!-- i18n:comment<NBSP>text -->`. It reaches `mdbook-gettext`/`mdbook-xgettext` through `group_events`.

Step past the separator by its UTF-8 width instead. The group_events fuzzer can reach this path but runs with `-only_ascii=1`, so it never generated the input.

### DIFF
--- a/i18n-helpers/src/directives.rs
+++ b/i18n-helpers/src/directives.rs
@@ -31,13 +31,17 @@ pub fn find(html: &str) -> Option<Directive> {
         (Some("skip"), Some("start")) => Some(Directive::SkipStart),
         (Some("skip"), Some("end")) => Some(Directive::SkipEnd),
         (Some("comment"), _) => {
-            let start_of_comment_offset = std::cmp::min(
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Very nice! Thanks for finding and fixing this.

Do we have the same problem with other directives?

sage-mode-hunter:
No, comment was the only one. The skip variants are decided purely by matching the whitespace-split tokens, so they never index back into the raw string. The comment branch was the only one slicing by a fixed byte offset to recover the trailing body, which is what let a multi-byte separator land mid-codepoint.

mgeisler:
Cool, thanks for checking!

PR #311, opened 2026-06-26
skip plural messages in add_stripped_summary_translations

sage-mode-hunter:
add_stripped_summary_translations unwraps msgstr() for every catalog entry sourced from SUMMARY.md.

- polib returns an error from msgstr() on plural messages rather than a string
- a translation PO with a plural entry sourced from SUMMARY.md then panics the gettext preprocessor mid-build
- the normal translate path already tolerates plural entries via msgstr().ok(), this one did not

A SUMMARY.md title is always singular, so skip entries whose msgstr() is an error instead of unwrapping. Added a regression test alongside the existing one.

### DIFF
--- a/i18n-helpers/src/gettext.rs
+++ b/i18n-helpers/src/gettext.rs
@@ -73,9 +73,16 @@ pub fn add_stripped_summary_translations(catalog: &mut Catalog) {
             continue;
         }
 
+        // SUMMARY.md holds singular chapter titles. A plural entry
+        // can't be one, and `msgstr` returns an error for plural
+        // messages, so skip it rather than unwrapping.
+        let Ok(msgstr) = msg.msgstr() else {
+            continue;
+        };
+
         let message = Message::build_singular()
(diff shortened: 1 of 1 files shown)


qwandor (approved):

qwandor:
This seems to causing a breakage in the fuzz tests: https://github.com/google/mdbook-i18n-helpers/actions/runs/28231315978/job/83635474937

sage-mode-hunter:
That failure is in the group_events target, not gettext. It panics inside pulldown-cmark (parse.rs:2199, Option::unwrap() on a None) reached through extract_events/group_events/reconstruct_markdown. This PR only touched add_stripped_summary_translations, which neither fuzz target calls: group_events fuzzes the markdown grouping path and the gettext target goes through translate_book. Looks like a separate parser crash the fuzzer happened to hit on the merge commit rather than anything from here. I can dig into the failing input if you want it tracked separately.

PR #312, opened 2026-06-28
avoid line-number overflow in compute_source

sage-mode-hunter:
**Integer overflow when offsetting source line numbers**
compute_source adds the per-paragraph delta back onto the line number parsed out of each `#:` source comment, and that number is taken verbatim from the input `.po`/`.pot`. A source like `foo.md:18446744073709551615` overflows the add once a multi-paragraph msgid is split into paragraphs that carry a non-zero offset, which panics mdbook-i18n-normalize under overflow checks (and wraps to a bogus line number otherwise). Switched it to saturating_add.


mgeisler (approved):
Thanks, that looks very nice!

Issue #78, opened 2023-09-13
Add to `install-action`


mgeisler:
It would be cool if one could install mdbook-i18n-helpers using [install-action](https://github.com/taiki-e/install-action).

miroim:
Hi @mgeisler , I am interested in working on this.
When I generate manifest the [guide](https://github.com/miroim/install-action/blob/main/DEVELOPMENT.md) from install-action,

mgeisler:
Hi @miroim, nice to meet you!

Thanks for looking into this. I agree we should generate binaries as part of the release — could you look into how people normally do that?

My Rust projects have mostly been libraries so far, and there nobody asked for binaries to be released 🙂

I use https://github.com/mgeisler/rust-release-gh-action for releasing my libraries and I would like to use it here too (eventually). Right now, the action requires a `lib.rs` file,p but that is something which should be fixed.

mgeisler:
> To make sure everything works as expected, I'll first conduct tests in the forked repository before integrating any changes into the main project.

That's the best way for this kind of thing! Then you can (force-)push as much as you want and get quick feedback!

Issue #48, opened 2023-08-14
Add a GitHub Action


mgeisler:
We should publish our own GitHub Action to make it easy to use `mdbook-i18n-helpers`.

This will move logic from the YAML file (such as [`publish.yml`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml)) to the action. The action is probably written in TypeScript (I'm not sure we can write an action in Rust?) and this will be a more comfortable language than what we can do with scripts in the YAML file.

mgeisler:
What should a GitHub action do?

- it should download a pre-compiled `mdbook-gettext` binary so that `mdbook build` can be fast.
- when #84 is done, the action should also download a pre-compiled version of the new binary there.

Where do these pre-compiled binaries come from? We will need to generate and publish them as part of a release workflow for this repository. See #78 for another issue about releasing binaries.

I would like us to use [mgeisler/rust-release-gh-action](https://github.com/mgeisler/rust-release-gh-action) to do the releases — it's the release flow I'm using for all my Rust crates these days. This release workflow needs to be integrated and it needs to be extended to generate binaries. It should probably also be rewritten as a Rust binary instead (which we will execute via a [Dockerfile GitHub action](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action).

So, all in all, there is a fair amount of steps to be tackled here by someone... it will involve touching several interesting technologies :smile:

trooper2013:
@mgeisler Could you please assign this task to me.

mgeisler:
Hey @trooper2013, thanks a lot for looking at this!

I would love for you to start simple here: building a fully automatic publishing pipeline is complicated, but making a very simple GitHub action is not.

See https://github.com/dprint/check as a good example of how to get started. The `action.yml` file there installs the `dprint` command line and runs it. In our case, "installing" means running `cargo install mdbook-i18n-helpers` at first :slightly_smiling_face:

carreter:
@trooper2013 are you still working on this? If not I can take a shot at it.

carreter:
@mgeisler If I'm understanding correctly, this action should do the following (modeled after (publish.yml)[https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml] from Comprehensive Rust):

- Install `mdbook`
- Install `mdbook-i18n-helpers` and dependencies
    - I know of `gettext`, anything else?
[...]

PR #313, opened 2026-06-30
CI: set up action to automatically create and build releases

orangecms:
see https://github.com/google/mdbook-i18n-helpers/issues/48#issuecomment-4847124077 / #78

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - mdbook-i18n-helpers-[0-9]+.*
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/create-gh-release-action@v1
+        with:
+          prefix: "mdbook-i18n-helpers-"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  upload-assets:
+    needs: create-release
+    strategy:
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/upload-rust-binary-action@v1
+        with:
+          bin: "mdbook-gettext,mdbook-i18n-normalize,mdbook-xgettext"
+          archive: $tag-$target
+          target: ${{ matrix.target }}
+          token: ${{ secrets.GITHUB_TOKEN }}


orangecms:
Took a few attempts, and got me here:
https://github.com/orangecms/mdbook-i18n-helpers/releases/tag/mdbook-i18n-helpers-0.4.7

That is:

```sh
tar -tf mdbook-i18n-helpers-0.4.7-x86_64-unknown-linux-gnu.tar.gz                                                                                                  9.656s
mdbook-gettext
mdbook-i18n-normalize
mdbook-xgettext
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+  push:
+    tags:
+      - mdbook-i18n-helpers-[0-9]+.*
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/create-gh-release-action@v1
```
qwandor, line 16: What advantage does this give over the official `actions/create-release`? How does it know where to find the changelog, given that we have two different crates in the workspace? Maybe it should just create a draft release, so a human can still review it before publishing it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - mdbook-i18n-helpers-[0-9]+.*
```
qwandor, line 9: What about the `i18n-report` crate which is also in the workspace?

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/checkout@v4
+      - uses: taiki-e/create-gh-release-action@v1
+        with:
+          prefix: "mdbook-i18n-helpers-"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  upload-assets:
+    needs: create-release
+    strategy:
+      matrix:
```
qwandor, line 24: What's the point of a matrix with only one entry?

orangecms: It's pretty much a copy of another CI setup I had, where someone asked for Windows binaries at some point, which could also be added here easily.

.github/workflows/release.yml
orangecms: I didn't consider it for now, would suggest creating a separate CI job.

.github/workflows/release.yml
orangecms: I haven't checked whether it even uses changelogs.
My goal here was really just to publish the binaries for a start.

orangecms: So I quickly checked - the changelog path can just be provided:
https://github.com/taiki-e/create-gh-release-action#example-workflow-basic-usage

I'll add that, and drafts are possible as well, see the other example below - I'll do that.

orangecms: Re https://github.com/actions/create-release :
That one has been obsolete for 5 years due to lack of maintenance. 😬
It's been archived since.

PR #320, opened 2026-07-04
guard po_file::parse against panics on malformed input

sage-mode-hunter:
polib panics instead of returning an error on some malformed PO lines (an unquoted `msgid` value, a bare `"` continuation), so a bad translation file aborts the whole run instead of failing gracefully.
- the gettext preprocessor and mdbook-i18n-normalize parse translator-supplied PO files and already `with_context` the call, expecting an error not a panic
- added `parse_catalog` to catch the panic and return that parse error instead
- routed the three `po_file::parse` sites (preprocessor, normalizer, i18n-report) through it

Repro: `printf 'msgid x\n' > bad.po; mdbook-i18n-normalize bad.po out.po` aborts before this, now reports a parse error.

### DIFF
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -47,6 +47,8 @@ pub mod xgettext;
 /// `msgid` line whose value is not wrapped in quotes) instead of returning an
 /// error. Guard the call so an invalid translation file is reported as a parse
 /// error rather than aborting the process.
+///
+/// Remove when https://github.com/BrettDong/polib/issues/25 is closed.
 pub fn parse_catalog(path: &std::path::Path) -> anyhow::Result<Catalog> {
     match std::panic::catch_unwind(|| polib::po_file::parse(path)) {
         Ok(result) => result.map_err(|err| anyhow::anyhow!("{err}")),
--- a/i18n-helpers/src/lib.rs
+++ b/i18n-helpers/src/lib.rs
@@ -41,6 +41,19 @@ pub mod preprocessors;
 pub mod renderers;
 pub mod xgettext;
 
+/// Parse the PO/POT file at `path` into a catalog.
+///
+/// `polib::po_file::parse` panics on some malformed input (for example a
+/// `msgid` line whose value is not wrapped in quotes) instead of returning an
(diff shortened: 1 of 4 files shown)


mgeisler:
Hi @sage-mode-hunter, thanks reporting this!

I would say that the correct place to fix this is in polib itself, right? We can of course be extra careful here until it is fixed upstream, but ideally we should export this kind of complexity to that library.

Can you file an issue upstream and reference it in this change? That will make it easier to know if/when we can remove it again.

Also, do we need a new fuzz target in our create to discover this kind of bug?

i18n-helpers/src/lib.rs
```diff
@@ -41,6 +41,19 @@ pub mod preprocessors;
 pub mod renderers;
 pub mod xgettext;
 
+/// Parse the PO/POT file at `path` into a catalog.
+///
+/// `polib::po_file::parse` panics on some malformed input (for example a
+/// `msgid` line whose value is not wrapped in quotes) instead of returning an
+/// error. Guard the call so an invalid translation file is reported as a parse
+/// error rather than aborting the process.
```
mgeisler, line 49: Basically, we need an issue to refer to here:

```suggestion
/// error rather than aborting the process.
///
/// Remove when https://github.com/BrettDong/polib/issues/NNN is closed.
```

mgeisler (changes requested):

i18n-helpers/src/lib.rs
sage-mode-hunter: Done, filed BrettDong/polib#25 and added the reference here and to the copy in i18n-report.

sage-mode-hunter:
Agreed that polib is the right place. Filed BrettDong/polib#25 (the slicing code is unchanged on master) and referenced it in the doc comments as suggested.

On the fuzz target: our current targets build catalogs in memory through the polib API, so the PO text parser is never exercised. A target feeding raw bytes into po_file::parse would have found this, but it would mostly be fuzzing polib and would sit red on this known crash until the upstream fix lands. It seems better placed upstream, close to what you proposed in BrettDong/polib#6. Happy to add one here once polib is fixed if you still want the coverage.

sage-mode-hunter:
any update?

